PR #3, opened 2016-10-25
[MCC-259892] Change MAuthSigningHandler options to MAuthSigningOptions

lschreck-mdsol:
This PR changes the used options instance in MAuthSigningHandler from MAuthOptions to MAuthSigningOptions. This change is necessary in order to not require MAuthServerUrl for the signing (which is not using this url at all).

Additionally this PR contains a fix in the OWIN middleware for the selfhosted request body being non-seekable. In these cases we replace the stream with a MemoryStream to be able to restore the body after reading it for authentication.

@mdsol/team-51 Please review and merge if it OK! Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ cases.
 ### Signing Outgoing Requests
 
 In order to sign outgoing requests, an `MAuthSigningHandler` class is provided in the Core package. this handler
-accepts an `MAuthOptions` instance which stores all the necessary settings for MAuth.
+accepts an `MAuthSigningOptions` instance which stores all the necessary settings for the signing process.
 
 An example:
 
@@ -91,7 +91,7 @@ using Medidata.MAuth.Core;
 
 public async Task<HttpResponseMessage> SignAndSendRequest(HttpRequestMessage request)
 {
-	var signingHandler = new MAuthSigningHandler(new MAuthOptions()
+	var signingHandler = new MAuthSigningHandler(new MAuthSigningOptions()
 	{
 		ApplicationUuid = new Guid("7c872d75-986b-4c61-bb17-f2569d42bfb0"),
 		PrivateKey = File.ReadAllText("ClientPrivateKey.pem")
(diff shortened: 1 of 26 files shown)


README.md
```diff
@@ -239,3 +239,16 @@ so in the near future there is a possibility that we will add a new package with
 
 We are using the latest version (as of date 1.81) of the [BouncyCastle](https://github.com/bcgit/bc-csharp) library.
 
+##### What are the major changes in the 2.0.0 version?
```
cabbott: Typically we have a CHANGELOG.md file in the root of the repository that tracks changes specifically. Here is an example in ruby client: https://github.com/mdsol/mauth-client-ruby/blob/master/CHANGELOG.md

lschreck-mdsol: Added the CHANGELOG.md with the brief summary of the changes. This section in the readme I thought more of an explanation of why these changes are needed and why we are jumping a major version (i.e. MAuthSigningHandler constructor signature change). I think this section can be removed in the next major version update.

lschreck-mdsol: Added the CHANGELOG.md with the brief summary of the changes (4e4fb5d). This section in the readme I thought more of an explanation of why these changes are needed and why we are jumping a major version (i.e. MAuthSigningHandler constructor signature change). I think this section can be removed in the next major version update.

cabbott:
LGTM 🎋

src/Medidata.MAuth.Owin/MAuthOwinExtensions.cs
```diff
@@ -54,5 +61,30 @@ public static HttpRequestMessage ToHttpRequestMessage(this IOwinRequest request)
                 throw;
             }
         }
+
+        public static async Task EnsureRequestBodyStreamSeekable(this IOwinContext context)
+        {
+            if (context.Request.Body == null || context.Request.Body == Stream.Null || context.Request.Body.CanSeek)
```
bvillanueva-mdsol: if (context.Request.Body.CanSeek)
Should it still undergo the process below? 
Since it will still be used later in request.Content.ReadAsStringAsync(), the pointer will be still set to last instead of zero before it enters the controller. Or is it reset somewhere to zero before entering the controller part?
What do you think @lschreck-mdsol ?

lschreck-mdsol: Yep, I added the rewind part to the ToHttpRequestMessage method:
https://github.com/mdsol/mauth-client-dotnet/pull/3/files#diff-5c82ef6a371cf02e30e4423ab4fb8bbbR32

Which will work only if the stream can seek. Maybe conceptually not the perfect place, but for now it will do the trick (the idea is to rewind the stream at the most closest point of the read operation).

bvillanueva-mdsol:
LGTM :+1: 
I think we can merge this. I'll merge later if no one objects. Thanks!

bvillanueva-mdsol:
Merging!

PR #12, opened 2017-02-27
[MCC-280769] Add retrying for the MAuth service communication

lschreck-mdsol:
This PR will add a retrying policy for the MAuth service request and also will decrease the default timeout from 10 seconds to 3 seconds.

@mdsol/team-51 Please review and merge if It's OK. Thanks!
@jcarres-mdsol I set the retry policy to `Normal` (3 overall attempts) - do you think it should be `Single`?


jcarres-mdsol:
I think eventually we will need to implement Retry Throttling, which is complex to implement as you need history of the past but the only way to avoid a bunch of problems in distributed systems.

Till then, we should consider all retries to be evil and do them only if absolutely necessary. Retrying once falls into this category as failing just because a small problem in the network is too strict but  I'd shy of retrying more than that.

In short, I favor `Single`.

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changes in Medidata.MAuth
 
-## v2.1.2
+## v2.2.0
+- **[Medidata.MAuth.Core]** Decreased the **default** timeout from 10 seconds to 3 seconds for the MAuth service
+requests in order to decrease the chance of service request congestion (the timout still configurable in the options)
```
jcarres-mdsol: timout -> timeout is

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
...
+- **[Medidata.MAuth.Core]** Added a new feature to make multiple attempts to communicate with the MAuth service in case
+if there are unsuccessful responses. The number of attempts (i.e. retry policy) is configurable through the options
```
jcarres-mdsol: You may not need `if` here

README.md
```diff
@@ -146,7 +147,8 @@ The middleware takes an `MAuthMiddlewareOptions` instance to set up the authenti
...
 | **PrivateKey** | Determines the RSA private key of the server application for the authentication requests. This key must be in a PEM ASN.1 format. |
-| **AuthenticateRequestTimeoutSeconds** | An optional parameter that determines the timeout in seconds for the MAuth authentication request - the MAuth component will try to reach the MAuth server for this duration before it throws an exception. If not specified, the default value will be **10 seconds**. |
+| **AuthenticateRequestTimeoutSeconds** | An optional parameter that determines the timeout in seconds for the MAuth authentication request - the MAuth component will try to reach the MAuth server for this duration before it throws an exception. If not specified, the default value will be **3 seconds**. |
+| **MAuthServiceRetryPolicy** | The policy for the retry attempts when communicating with the MAuth service. The following policies can be used: `NoRetry` (no retries), `Single` (one additional attempt), `Normal` (two additional attempts) and `Agressive` (9 additional attempts) - the default value is **Normal**. |
```
jcarres-mdsol: I'd say `Normal` is `Multiple` or `Twice`. Difficult to know what's Normal

src/Medidata.MAuth.Core/MAuthRequestRetrier.cs
```diff
@@ -0,0 +1,65 @@
...
+                    "No request function provided or the provided request function resulted null request.",
+                    nameof(requestFactory)
+                );
+
+            exception = exception ?? new RetriedRequestException(
+                $"Could not get a successful response from the MAuth Service after {remainingAttempts} attempts. " +
+                "Please see the responses for each attempt in the exception's Responses field.");
+
+            if (remainingAttempts == 0)
+                throw exception;
```
hkurniawan-mdsol: I think we can refactor `exception = exception ?? new RetriedRequestException( ` to something like below.
```cs
if (remainingAttempts == 0) 
  new RetriedRequestException( 
    $"Could not get a successful response from the MAuth Service after {noOfAttempts} attempts. " + 
    "Please see the responses for each attempt in the exception's Responses field."); 
```
and have `noOfAttempts` to be assigned in the constructor.

What do you think?

src/Medidata.MAuth.Core/Exceptions/RetriedRequestException.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        /// <summary>
+        /// Determines the responses for each request attempts made.
+        /// </summary>
+        public ICollection<HttpResponseMessage> Responses { get; } = new List<HttpResponseMessage>();
+
+        /// <summary>
+        /// Determines the request message of the first request attempt to the MAuth server.
+        /// </summary>
+        public HttpRequestMessage Request { get; set; }
```
hkurniawan-mdsol: Could you tell me where this is being used?

src/Medidata.MAuth.Core/MAuthRequestRetrier.cs
lschreck-mdsol: If the `remainingAttempts` is 0 it is too late to create the exception as we store the previous responses in the exception's `Responses` property. So this exception has to exist at this point (and it will be thrown here).

The `exception = exception ?? new RetriedRequestException()` is a shorter way for a null check (the runtime won't assign the `exception` to itself if it is not null).

src/Medidata.MAuth.Core/Exceptions/RetriedRequestException.cs
lschreck-mdsol: This is for informational purposes. The responses and the request itself is stored inside the exception in order to help the debugging/analysis.

README.md
lschreck-mdsol: You have a good point here, I tried to choose more sensible names: a5d1d5a

src/Medidata.MAuth.Core/MAuthRequestRetrier.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Linq;
+#if !NETSTANDARD1_4
+using System.Net.Cache;
+#endif
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Medidata.MAuth.Core
+{
+    internal class MAuthRequestRetrier
```
kenyamat: I thought a solution is to create RetryHandler implemented DelegatingHandler like the [article](http://stackoverflow.com/questions/19260060/retrying-httpclient-unsuccessful-requests).
then inject it into HttpClient on original `GetApplicationInfo` method. It will be smaller changes.

lschreck-mdsol: According to the linked SO answer (in the comments) it has some drawbacks, like not working with timeouts (when the server does not respond to you at all), so I would not implement it here. This current solution will give a good protection against minor network glitches I think.

src/Medidata.MAuth.Core/Exceptions/RetriedRequestException.cs
hkurniawan-mdsol: Does that mean there has to be a test on this as well?

src/Medidata.MAuth.Core/MAuthRequestRetrier.cs
hkurniawan-mdsol: Yes, I do know what `??` operator does.

I just think having the line `exception = exception ?? new RetriedRequestException(` to be executed every time where it actually only used inside the `if (remainingAttempts == 0)` is a bit strange.

The `exception.Responses` objects can be easily resolved by declare a class level property
```cs
internal class MAuthRequestRetrier
{
  private ICollection<HttpResponseMessage> responseResults = new List<HttpResponseMessage>()
  ...
}
```
Then
```cs
if (remainingAttempts == 0) {
  var retriedException = new RetriedRequestException(...);
  retriedException.Responses = responseResults;
  throw retriedException;
}
```

lschreck-mdsol: So you mean that instead of having a simple assignment (which will be skipped by the runtime every but the first call either way) with a coalescing operator is simpler to have the following:
- A multiline `if` block with the same operation as the coalescing operator's right side
- A dedicated constructor parameter only for the sole purpose of putting it to the exception message
- An additional private List field just for the purpose of storing all the responses which we will assign to the exception either way in the end

In my humble opinion that is a lot of additional work/resources for the same result.

Also
> to be executed every time where it actually only used inside the `if (remainingAttempts == 0)`

is not true, as it won't be executed every time, only the first time: when the exception is null (in this case it will instantiate a new `RetriedRequestException`). Every other time it will be `exception = exception` which will be skipped by the runtime.

src/Medidata.MAuth.Core/Exceptions/RetriedRequestException.cs
lschreck-mdsol: I don't think that this needs a test - it is not used at runtime and I realized my mistake already so a test would not add any real value here.

src/Medidata.MAuth.Core/MAuthRequestRetrier.cs
hkurniawan-mdsol: - You can throw the exception inside the `if` statement in shorter form if you want to.
`throw new RetriedRequestException() { ... }`

- There's no need for a dedicated constructor.
You already have `MAuthOptionsBase` inside `MAuthRequestRetrier`.
What you need is add a property like  
`public int NoOfAttempts { get; set; } = (int)MAuthServiceRetryPolicy + 1;`
to `MAuthOptionsBase` and use this wherever is needed.

If you want to keep this as is, that's fine.
However, is there a way not to instantiate `RetriedRequestException` on successful requests which should be 99% of the time.

PR #15, opened 2017-06-12
[MCC-306345] Add ASP.NET Core middleware and other enhancements

lschreck-mdsol:
This PR adds the following features to the MAuth library:

- Convert the projects to the new csproj format
- Add the ASP.NET Core middleware (based on the OWIN middleware)
- Make the private key format line ending agnostic
- Detect and load the private key if the key is actually a file path

Also because now the test project multitargets net452 and netcoreapp1.1, in the Visual Studio test runner only the net452 tests will show up (because currently it is not capable to show multitargeting tests). The command line xunit test runner will run both types of tests without any additional configuration or problem.

@mdsol/team-51 Please review/merge - thanks!


lschreck-mdsol:
Please ignore the failing checks, it was an experiment with AppVeyor but I forgot to remove the build (and now I cannot remove the result of those failed builds).

src/Medidata.MAuth.AspNetCore/MAuthMiddlewareOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// Contains the options used by the <see cref="MAuthMiddleware"/>.
+    /// </summary>
+    public class MAuthMiddlewareOptions : MAuthOptionsBase
+    {
+        /// <summary>
+        /// Determines if the middleware should swallow all exceptions and return an empty HTTP response with a
+        /// status code Forbidden (403) in case of any errors (including authentication and validation errors).
+        /// The default is <see langword="true"/>.
+        /// </summary>
+        public bool HideExceptionsAndReturnForbidden { get; set; } = true;
```
bvillanueva-mdsol: Looks like this can be transferred to `MAuthOptionsBase` class. This is also used in other derived classes. @lschreck-mdsol What do you think?

README.md
```diff
@@ -112,65 +122,91 @@ The `MAuthSigningOptions` has the following properties to determine the required
...
+    public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
+    {
+        app.UseMAuthAuthentication(options =>
+        {
+            options.ApplicationUuid = new Guid("a419de8f-d759-4db9-b9a7-c2cd14174987");
+            options.MAuthServiceUrl = new Uri("https://mauth.imedidata.com");
+            options.AuthenticateRequestTimeoutSeconds = 3;
+            options.MAuthServiceRetryPolicy = MAuthServiceRetryPolicy.RetryOnce;
+            options.HideExceptionsAndReturnForbidden = true;
+            options.PrivateKey = "ServerPrivateKey.pem";
```
bvillanueva-mdsol: Should this Private key value in absolute path format?

bvillanueva-mdsol:
Please check 2 warnings in solution. Both warnings points to mismatch in comment description of parameter and actual parameter name

tests/Medidata.MAuth.Tests/MAuthAspNetCoreTests.cs
```diff
@@ -0,0 +1,151 @@
+﻿using System.IO;
```
bvillanueva-mdsol: @lschreck-mdsol This file is not included inside the project. Please confirm. Thanks

bvillanueva-mdsol: Or maybe my visual studio is weird

src/Medidata.MAuth.AspNetCore/MAuthMiddlewareOptions.cs
lschreck-mdsol: Technically you are right, but architecturally IMHO this should be handled on the framework (OWIN, WebAPI and AspNetCore) component level, because the Core does not care of what will happen with the exceptions.

Although currently all the framework components are able to hide or show exceptions, this is not guaranteed to be true for any future implementations. Also for example in the tests with the `MAuthTestOptions` we don't need this property as we don't use it during the testing; but if we add it to the `MAuthOptionsBase`, it will be exposed for the tests as well causing some unnecessary noise on the options interface.

README.md
lschreck-mdsol: It can be either relative or absolute path (UNC-paths also should work), in this example I used relative path (originally this example was `File.ReadAllText()` with a relative path so this is not changed after all).

src/Medidata.MAuth.AspNetCore/MAuthMiddlewareOptions.cs
bvillanueva-mdsol: Got it. Not being used in core then. Im ok this being at the derived class, it would be crazy too if we put another level of option class between base and current derived class if it only houses this property.

tests/Medidata.MAuth.Tests/MAuthAspNetCoreTests.cs
lschreck-mdsol: Ok, this is the tricky part. It is actually included when you compile the project targeting aspnetcore1.1 😉 . It is just Visual Studio does not show you as included because by default it shows the solution as it targets the first target framework which is net452.

You can tell if this is included by running `dotnet xunit` in the Package Manager Console from the `.\tests\Medidata.MAuth.Tests` folder, and it should run and show you both framework tests (71 for net452 and 59 for netcoreapp1.1 because not every test can be run in both frameworks).

PR #22, opened 2017-07-27
[MCC-319001] Add a Utility extension class to support calling MAuth methods

lschreck-mdsol:
This PR adds a public utility extension class which contains some methods to call outside of the MAuth context. This can be useful in scenarios where you have to work with MAuth specific data (headers for example) and don't want to reimplement the function in your own code.

Later on this class will be extended to have more methods and ultimately the internal MAuth implementation will start to use these methods instead of the internal implementations (before you ask: of course the current internal implementations cannot be copied to this public class as they are because they have signatures and/or they are using internal data structures which do not make sense to expose).

@mdsol/team-51 Please review/merge, then I will send a new PR to have a new client version out - Thanks!

### DIFF
--- a/src/Medidata.MAuth.Core/UtilityExtensions.cs
+++ b/src/Medidata.MAuth.Core/UtilityExtensions.cs
@@ -1,0 +1,59 @@
+﻿using System;
+
+namespace Medidata.MAuth.Core
+{
+    /// <summary>
+    /// Contains utility extension methods for MAuth.
+    /// </summary>
+    public static class UtilityExtensions
+    {
+        /// <summary>
+        /// Parses an MAuth authentication HTTP header value for the application uuid and the MAuth signature payload.
+        /// If the parsing is not successful, an <see cref="ArgumentException"/> is thrown.
+        /// </summary>
+        /// <param name="headerValue">
+        /// The value of the X-MWS-Authentication HTTP header containing the application uuid and the MAuth signature
+        /// payload.
+        /// </param>
+        /// <returns>The application uuid and the payload encoded with Base64 encoding.</returns>
+        public static (Guid Uuid, string Base64Payload) ParseAuthenticationHeader(this string headerValue) =>
+            headerValue.TryParseAuthenticationHeader(out var result) ?
+                result :
+                throw new ArgumentException("Invalid MAuth authentication header value.", nameof(headerValue));
+
+        /// <summary>
+        /// Parses an MAuth authentication HTTP header value for the application uuid and the MAuth signature payload.
+        /// A return value indicates whether the parsing succeeded.
+        /// </summary>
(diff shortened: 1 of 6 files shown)


src/Medidata.MAuth.Core/UtilityExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <returns>
+        /// <see langword="true"/> if <paramref name="headerValue"/> was parsed successfully; otherwise,
+        /// <see langword="false"/>.
+        /// </returns>
+        public static bool TryParseAuthenticationHeader(this string headerValue,
+            out (Guid Uuid, string Base64Payload) result)
+        {
+            var match = Constants.AuthenticationHeaderRegex.Match(headerValue);
+
+            result = default((Guid, string));
```
hkurniawan-mdsol: Is setting default value here necessary?
If there is no match then `ArgumentException` will be thrown and if there is, result will contain `uuid` and `payload`.

hkurniawan-mdsol: Aah, I think I know why. The `out` parameter needs to be assigned inside `TryParseAuthenticationHeader` right?

lschreck-mdsol: Yes, it is necessary, if you don't set the value of an `out` argument before the control leaves the current method, you will get a compiler error ([CS0177](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0177)).

Also this is in the `TryParseAuthenticationHeader()` method which won't throw exception in case of no match.

kenyamat:
any unit tests?

lschreck-mdsol:
@kenyamat No meaningful code change here, just reorganizing + the existing unit tests cover these methods as well (`GetAuthenticationInfo_WithSignedRequest_WillReturnCorrectAuthInfo()` for example).

kenyamat:
@lschreck-mdsol I recommend to reorganize or create unit tests for code on UtilityExtensions into UtilityExtensionsTests.cs for maintainability.

lschreck-mdsol:
@kenyamat Are you sure that I should spend more development time writing redundant and unnecessary tests which possibly will just increase the maintainability costs (now you have to maintain another test file with another test class and another test methods)?

kenyamat:
@lschreck-mdsol Yes, I don't think this is redundant time or unnecessary tests for maintainability. As you know we often change owner or developer for our own product. And this is open source product.
These unit tests to cover all path is very important for future refactoring. I don't think I need to describe unit test importance. Anyway I will created the unit tests if you think it's redundant time.

lschreck-mdsol:
@kenyamat Understood, let me add those unit tests, then.

kenyamat:
Thanks.

PR #24, opened 2017-08-31
[MCC-328941] Fix request body content not rewound after authentication

lschreck-mdsol:
This PR resolves the issue with the incoming request body stream being not rewound after authentication.

The stream will be rewound at the end of the ASP.NET Core and OWIN middlewares to ensure the next middleware gets the stream rewound down the chain.

@mdsol/team-51 Please review/merge. Thanks!

### DIFF
--- a/src/Medidata.MAuth.AspNetCore/MAuthAspNetCoreExtensions.cs
+++ b/src/Medidata.MAuth.AspNetCore/MAuthAspNetCoreExtensions.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Medidata.MAuth.Core;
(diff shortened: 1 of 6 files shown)


hkurniawan-mdsol:
Is there a test to make sure that the stream is rewound so the next middleware can read?

lschreck-mdsol:
@hkurniawan-mdsol I did an adhoc with SLoginator.

src/Medidata.MAuth.AspNetCore/MAuthAspNetCoreExtensions.cs
```diff
@@ -18,17 +17,18 @@ internal static class MAuthAspNetCoreExtensions
...
 
             foreach (var header in request.Headers)
             {
                 if (!result.Headers.TryAddWithoutValidation(header.Key, (IEnumerable<string>)header.Value))
                     result.Content.Headers.TryAddWithoutValidation(header.Key, (IEnumerable<string>)header.Value);
             }
 
-            if (request.Body.CanSeek)
-                request.Body.Seek(0, SeekOrigin.Begin);
+            request.Body.Rewind();
```
kenyamat: Can we remove the rewind? The followings are the call stack but request.Body has not been read until here.

MAuthMiddleware.Invoke -> MAuthAspNetCoreExtensions.TryAuthenticate -> MAuthAspNetCoreExtensions.ToHttpRequestMessage

lschreck-mdsol: Removed the Rewind from here and relocated for the Owin middleware.

For the record, in case of ASP.NET Core we have the built in `EnableRewind()` method for request body, but for OWIN we have to create a memory stream instead which was not rewind.

kenyamat:
Looks good. It worked on my local sLog.

PR #29, opened 2018-01-31
[MCC-365432] Change uuid validation to accept non-v4 uuids

lschreck-mdsol:
As the title says, this PR will remove the restriction on the MAuth uuid validation to be only version 4.

@mdsol/team-51 Please review and merge - then we can release it to the wild! Thanks!


hkurniawan-mdsol:
Should we add a test in [UtilityExtensionsTest.cs](https://github.com/mdsol/mauth-client-dotnet/blob/hotfix/non-v4-uuids/tests/Medidata.MAuth.Tests/UtilityExtensionsTest.cs) to check for v1 uuid?

bvillanueva-mdsol:
+1 for additional test.

PR #31, opened 2018-06-05
[MCC-400374] Allow MAuthServerHandler injection from client code

jportaro-mdsol:
Currently, the mAuth .NET client leverages `WebRequestHandler` to handle HTTP caching.  Unfortunately, the `WebRequestHandler` implementation of caching isn't supported on IIS.  As such, in our use case, even with cache headers set correctly the mAuth will round-trip to the mAuth server to get the public key.  

This PR exposes the `MAuthOptionsBase.MAuthServerHandler` as a public property such that server friendly caching can be implemented.


hkurniawan-mdsol:
A couple of things I'd like bring up regarding CacheCow's [CachingHandler](https://github.com/aliostad/CacheCow/blob/9f0e46cc86e98d2def5c064b6d0761c04e237fb7/src/CacheCow.Client/CachingHandler.cs#L18).
- Looking at the [SendAsync()](https://github.com/aliostad/CacheCow/blob/9f0e46cc86e98d2def5c064b6d0761c04e237fb7/src/CacheCow.Client/CachingHandler.cs#L254), I found it hard to follow. Not sure whether this merits a concern
- I'm assuming you'll be using the [InMemoryCacheStore](https://github.com/aliostad/CacheCow/blob/3d15b138ccd58b5bc27425f14f57e4b9d2d0402e/src/CacheCow.Client/InMemoryCacheStore.cs). It calls [DeserializeToResponseAsync](https://github.com/aliostad/CacheCow/blob/b351bdefa49700e780ce576b8347358b7a701586/src/CacheCow.Client/MessageContentHttpMessageSerializer.cs#L70) and in there `"Content-Type", "application/http;msgtype=response"` header is added. Will this negatively affect your end?

I'm generally fine with this minor change. If it is necessary for your app/service to have this caching feature, then let's go ahead with this.
Otherwise as @lschreck-mdsol mentioned, I'd rather we do it proper.

jportaro-mdsol:
@hkurniawan-mdsol Thanks for the notes.  

- Agreed - The code isn't exactly an easy read.  However, it's being used in a very focused code path hopefully for a limited time.  Additionally, despite the complexity, it appears to be handling the cache header correctly.  As such, I don't expect to have I'll need to make any changes.  
- Assumptions are correct, and it doesn't appear that will be an issue.  After taking a closer look, it seems that the header is there to meet some preconditions of the `ReadAsHttpResponseMessageAsync` call.  When the cached response finally bubbles up to the mAuth code, `"Content-Type", "application/http;msgtype=response"` isn't even in the header collection.

hkurniawan-mdsol:
@jportaro-mdsol 
Oh the `application/http;msgtype=response` didn't bubble up!? Okay, no worries then.

jportaro-mdsol:
Yes, it's a really weird de-serialization pattern.  Again, thanks for the notes.  It's much better understood now.

PR #34, opened 2018-08-21
[MCC-422186] Fix concurrency issue in MAuthRequestRetrier

johnduhart:
The `exception` field in `MAuthRequestRetrier` can be modified from multiple threads, additionally even when an exception would be thrown, the `Request` property would likely be wrong, and the `Responses` contained responses from other failed & successful requests.


hkurniawan-mdsol:
Thank you for the bug fix @johnduhart!

PR #36, opened 2018-10-03
[MCC-433771] Fix MAuth authentication with Binary HTTP content

lschreck-mdsol:
This PR attempts to fix the authentication in case of binary HTTP content: we will construct the signature with binary concatenating.

Additionally because the unit test infrastructure was not able to handle binary content, I refactored it a little bit.

@mdsol/team-51 Please review/merge - thanks!


tests/Medidata.MAuth.Tests/MAuthCoreExtensionsTests.cs
```diff
@@ -60,49 +60,68 @@ public async Task Verify_WithCorrectlySignedData_WillVerifyTheDataAsValid()
...
-        public async Task GetSignature_WithRequest_WillReturnTheCorrectSignature(string method)
+        public static async Task GetSignature_WithRequest_WillReturnTheCorrectSignature(string method)
         {
             // Arrange
-            var testData = await TestData.For(method);
+            var testData = await method.FromResource();
 
             var expectedSignature =
-                ($"{method}\n/\n{testData.Content}\n{TestExtensions.ClientUuid.ToHyphenString()}\n" +
+                ($"{method}\n/\n{testData.Base64Content.ToStringContent()}\n" +
```
kenyamat: How about like this? It is easy to understand that signature includes absolute path.
```
($"{testData.Method}\n" + // or $"{method}\n" +
 $"{testData.Url.AbsolutePath}/\n" +
 $"{testData.Base64Content.ToStringContent()}\n" +
```

lschreck-mdsol: Good call! bd78b06

lschreck-mdsol:
Yo, @mdsol/team-51 Can you check and merge! It would be highly appreciated! 🎃

kenyamat:
LGTM

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -73,11 +74,15 @@ public static Task<bool> Verify(this byte[] signedData, byte[] signature, string
...
-            ($"{request.Method.Method}\n" +
-            $"{request.RequestUri.AbsolutePath}\n" +
-            $"{(request.Content != null ? await request.Content.ReadAsStringAsync() : string.Empty)}\n" +
-            $"{authInfo.ApplicationUuid.ToHyphenString()}\n" +
-            $"{authInfo.SignedTime.ToUnixTimeSeconds()}")
+            new byte[][]
+            {
+                request.Method.Method.ToBytes(), NewLine,
+                request.RequestUri.AbsolutePath.ToBytes(), NewLine,
+                (request.Content != null ? await request.Content.ReadAsByteArrayAsync() : new byte[] { }), NewLine,
```
bvillanueva-mdsol: Suggestion: Avoid unnecessary zero-length array allocations.  Use Array.Empty<byte>() instead.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -209,6 +214,8 @@ public static TValue GetFirstValueOrDefault<TValue>(this HttpHeaders headers, st
         public static byte[] AsSHA512Hash(this string value) =>
```
bvillanueva-mdsol: This one is only used by test. Should the two AsSHA512Hash methods be merged? 
Like
```
        public static byte[] AsSHA512Hash(this string value) =>
            AsSHA512Hash(Encoding.UTF8.GetBytes(value)));

        public static byte[] AsSHA512Hash(this byte[] value) =>
            Hex.Encode(SHA512.Create().ComputeHash(value));
```

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -124,8 +129,8 @@ public static PayloadAuthenticationInfo GetAuthenticationInfo(this HttpRequestMe
                 $"MWS {authInfo.ApplicationUuid.ToHyphenString()}:" +
                 $"{await request.CalculatePayload(authInfo)}";
 
-            request.Headers.Add(Constants.MAuthHeaderKey, authHeader);
-            request.Headers.Add(Constants.MAuthTimeHeaderKey, authInfo.SignedTime.ToUnixTimeSeconds().ToString());
+            request.Headers.Add(MAuthHeaderKey, authHeader);
```
bvillanueva-mdsol: This is just my opinion: having `Constants` before constant variable names improves readability. I can know immediately as a code reviewer that this is a constant.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
lschreck-mdsol: Can't 😢 - For `net452` there is no `.Empty()` method on the array...

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
lschreck-mdsol: Can you elaborate on what you mean by "merge methods"?

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
lschreck-mdsol: Makes sense 69dac85

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
bvillanueva-mdsol: I mean method `public static byte[] AsSHA512Hash(this string value)` should use `public static byte[] AsSHA512Hash(this byte[] value)` so that unit test is still using the logic being used by the code proper.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
bvillanueva-mdsol: Yes you are right. Lol. Nvm. Thanks!

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
lschreck-mdsol: Oh, you provided the solution code not the current code! Ok, totally makes sense, thank you! 44be350

tests/Medidata.MAuth.Tests/MAuthCoreExtensionsTests.cs
```diff
@@ -60,49 +60,70 @@ public async Task Verify_WithCorrectlySignedData_WillVerifyTheDataAsValid()
...
+            var testData = await method.FromResource();
 
             var expectedSignature =
-                ($"{method}\n/\n{testData.Content}\n{TestExtensions.ClientUuid.ToHyphenString()}\n" +
-                 $"{testData.SignedTime.ToUnixTimeSeconds()}")
+                ($"{testData.Method}\n" +
+                 $"{testData.Url.AbsolutePath}\n" +
+                 $"{testData.Base64Content.ToStringContent()}\n" +
+                 $"{testData.ApplicationUuidString}\n" +
+                 $"{testData.SignedTimeUnixSeconds}")
```
hkurniawan-mdsol: I see that this expected signature is constructed from string concatenation which is what mauth-client-dotnet is currently doing. 
If string concatenation works, why are we concatenating bytes?

tests/Medidata.MAuth.Tests/MAuthAuthenticatorTests.cs
```diff
@@ -8,49 +8,46 @@
 
 namespace Medidata.MAuth.Tests
 {
-    public class MAuthAuthenticatorTests
+    public static class MAuthAuthenticatorTests
```
hkurniawan-mdsol: What's the reasoning for having the test class and methods as static?

tests/Medidata.MAuth.Tests/MAuthCoreExtensionsTests.cs
lschreck-mdsol: That's the whole point of this PR and the past days discussion: it works only for string content, but not for binary content.

This test ensures that even with changing the signature construction to byte concatenation it will still work for string contents (that is, it's backward compatible).

tests/Medidata.MAuth.Tests/MAuthAuthenticatorTests.cs
lschreck-mdsol: The code analysis told me to do it (actually I don't think it is significant, but also there are some micro performance gains here).

tests/Medidata.MAuth.Tests/MAuthCoreExtensionsTests.cs
hkurniawan-mdsol: Doesn't the test contain binary content as well?
`$"{testData.Base64Content.ToStringContent()}\n"`

The only place needs to be changed on the concatenation is the way we read the payload I think.
Instead of
`await request.Content.ReadAsStringAsync()`
Change it to
`Encoding.UTF8.GetString(await request.Content.ReadAsByteArrayAsync())`

lschreck-mdsol: No, that's just the way how the test data is stored (I don't apply different storing (i.e. plain text vs Base64 encoded text) methods for test data).

Also if you decode back the binary data into a UTF-8 string, that's not improving the situation at all.

To be honest, the solution is already working, it is adhoced, it is unit tested, we don't need an alternative solution to resolve this problem - we just need this to be merged and released. That is all.

hkurniawan-mdsol: I am in no way suggesting alternative solution and I am completely aware that the solution is working.

I just saw a test that's doing string concatenation with some `payload` which what the old solution is doing except the only difference is reading the payload itself.

Anyways will merge thanks.

PR #41, opened 2018-10-16
[MCC-425214] use ConfigureAwait(false) 

kenyamat:
@mdsol/team-51 use ConfigureAwait(false) avoiding any deadlock on synchronous way. Please review and merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changes in Medidata.MAuth
 
+## v3.1.1
+- **[Core]** Added `ConfigureAwait(false)` avoiding any possible deadlocks.
+
 ## v3.1.0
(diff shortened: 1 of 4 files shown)


bvillanueva-mdsol:
Do we want to handle also non-"Core" code for using `ConfigureAwait(false)` ? 
Example: https://github.com/mdsol/mauth-client-dotnet/blob/6dd15af1009bde9b9751a7d3a59840a1c2f2e2b2/src/Medidata.MAuth.Owin/MAuthOwinExtensions.cs#L73
@kenyamat @lschreck-mdsol

kenyamat:
@bvillanueva-mdsol no we don't, this is special case for past implementation, web apps using core should be implement asynchronous properly.

PR #43, opened 2019-01-18
[MCC-465787] Add caching for the MAuth server communication

lschreck-mdsol:
This PR contains the caching mechanism for MAuth authentication that involves requests to the MAuth server.

The solution sets the expiration time based on the MAuth server response cache header (if it is not set, it will be 1 hour by default).

I will update the documentation as well.

### DIFF
--- a/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
+++ b/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Net.Http;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Caching.Memory;
 using Org.BouncyCastle.Crypto;
 
 namespace Medidata.MAuth.Core
@@ -8,7 +9,8 @@ namespace Medidata.MAuth.Core
     internal class MAuthAuthenticator
     {
         private readonly MAuthOptionsBase options;
-        private MAuthRequestRetrier retrier;
+        private readonly MAuthRequestRetrier retrier;
+        private readonly IMemoryCache cache = new MemoryCache(new MemoryCacheOptions());
 
         public Guid ApplicationUuid => options.ApplicationUuid;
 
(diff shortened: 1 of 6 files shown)


src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -60,11 +62,24 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
             }
         }
 
-        private async Task<ApplicationInfo> GetApplicationInfo(Guid applicationUuid) =>
-            await (await retrier.GetSuccessfulResponse(applicationUuid, CreateRequest,
-                requestAttempts: (int)options.MAuthServiceRetryPolicy + 1))
-            .Content
-            .FromResponse();
+        private Task<ApplicationInfo> GetApplicationInfo(Guid applicationUuid) =>
```
kenyamat: do you have any reason to remove async?

lschreck-mdsol: As the method body is just a single async method call without any awaiting, we can propagate the result `Task` to the caller - no need to async/await it here.

kenyamat: Oh, all awaits are on another method body. Thanks.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -60,11 +62,24 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
...
+            {
+                var response = await retrier.GetSuccessfulResponse(
+                    applicationUuid,
+                    CreateRequest,
+                    requestAttempts: (int)options.MAuthServiceRetryPolicy + 1
+                );
+
+                var result = await response.Content.FromResponse();
+
+                entry.SetOptions(
```
kenyamat: will auth error case be cached?

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -60,11 +62,24 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
...
+                    new MemoryCacheEntryOptions()
+                        .SetAbsoluteExpiration(response.Headers.CacheControl?.MaxAge ?? TimeSpan.FromHours(1))
```
hkurniawan-mdsol: Is there a reason for `SetAbsoluteExpiration` instead of `SetSlidingExpiration`?

lschreck-mdsol: Because the max age cache-control is relative to the request time: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Expiration

lschreck-mdsol:
@kenyamat FYI adhoced with invalid app uuids, it's not caching anything and calls the MAuth server continuously.

PR #45, opened 2019-07-16
[MCC-422164] Upgrade mauth client dotnet to use MWSV2 signature

prajon84:
This PR is to support MAuth client for MWSV2 signature. PR is created before all the tasks under this PR are completed so that the reviewers can start reviewing and provide feedback.

- [x] Added necessary MWSV2 signature flow.

- [x] Updated netstandard framework and aspnetcore framework.

- [x] Unit tests for the added V2 Signature class methods.

- [x] Update `version.props` and `CHANGELOG.md`

- [x] Update `README.md` 

Note: Loggin will be done in follow-up PR.

Please review. 
@mdsol/team-51 
@ejennings-mdsol 
@johnduhart 
@jcarres-mdsol


cmcinnes-mdsol:
Adding @mdsol/team-16 for visibility here.

prajon84:
Can you also please tell me the reason why "AppVeyor" build is failing too even though I have added the nuget package ?

bvillanueva-mdsol:
As far as I know, `mauth-client-dotnet` is an open source library. So all dependecies should be open source. `Medidata.Platform.Logging` should not be used as it is an internal library. Thus the error in appveyor. What is the reason for including `Medidata.Platform.Logging`? 

image

hkurniawan-mdsol:
Echoing @bvillanueva-mdsol, I don't think we can use `Medidata.Platform.Logging` for an OpenSource project. Also I believe we try not to log in a library. We leave that responsibility to the consuming application.

If we do have to then
- We need to create a mechanism where consuming application can inject a logger
- The library just pick an OpenSource logging library ([Serilog](https://github.com/serilog/serilog)) and log to a file. Serilog is a logging library and it has it's own logging when it encounters an error https://github.com/serilog/serilog/blob/8ed3afd0f2d942429832b6674a2f599c7fb1fa38/src/Serilog/Configuration/LoggerSinkConfiguration.cs#L80

prajon84:
> @bvillanueva-mdsol @hkurniawan-mdsol : Thank you for the answer. But, even though `Medidata.Platform.Logging` is removed; AppVeyor build still fails saying some unit test fails. However, no existing unit tests are failing locally.

Besides, logging is required based on the requirement. 
image

If  the way you mentioned above is the way to move forward, then I will start implementing in that way a) injecting logger b) using `Serilog` in the library.

johnduhart:
> b) using Serilog in the library.

Let's not do that. Can we look at either using [DiagnosticSource](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/DiagnosticSourceUsersGuide.md) for the library, or simply taking a dependency on MEL such that a logger implementation can be provided by the user?

src/Medidata.MAuth.AspNetCore/MAuthAspNetCoreExtensions.cs
```diff
@@ -39,16 +41,20 @@ public static HttpRequestMessage ToHttpRequestMessage(this HttpRequest request)
...
         /// otherwise it will return either <see langword="false"/> if the method should ignore exceptions or
         /// will throw an exception if any errors occurred during the authentication.
         /// </returns>
         public static async Task<bool> TryAuthenticate(
-            this HttpContext context, MAuthAuthenticator authenticator, bool shouldIgnoreExceptions)
+            this HttpContext context, MAuthAuthenticator authenticator, bool shouldIgnoreExceptions, Enum mAuthVersion)
         {
             try
             {
+                if (mAuthVersion.Equals(Version.V2))
```
johnduhart: This doesn't seem like the right approach, the client library should automatically opt to use the newer headers if they're available with an option to eventually reject V1 headers.

src/Medidata.MAuth.Core/MAuthCoreExtensionsV2.cs
```diff
@@ -0,0 +1,246 @@
...
+        /// </summary>
+        /// <param name="request">
+        /// The request which has the information (method, url and content) for the calculation.
+        /// </param>
+        /// <param name="authInfo">
+        /// The <see cref="PrivateKeyAuthenticationInfo"/> which holds the application uuid, the time of the
+        /// signature and the private key.
+        /// </param>
+        /// <returns>A task object which will result the payload as a Base64 encoded string when completed.</returns>
+        public static async Task<string> CalculatePayloadV2(
```
johnduhart: I'm really against having V2 versions of every method, and it's sadly a result of making _everything_ in this repository an extension method for some reason. Can we look have creating an Interface with singleton implementations for the two versions?

prajon84: Thank you John. I will create interface but just do not want to mess up older process much. However, If I create interface and remove these extensions method, I might need to refactor the older extension method too.

src/Medidata.MAuth.AspNetCore/MAuthAspNetCoreExtensions.cs
prajon84: Well, if in that case if the client should use newer header if available then, how to maintain backwards compatibility ? little confused here..
image

johnduhart: You maintain backwards compatibility by looking to see if the newer header is present, and then consuming it if so.

prajon84: Oh Yes. It's not necessary for authenticate. Thank you

prajon84:
> > b) using Serilog in the library.
> 
> Let's not do that. Can we look at either using [DiagnosticSource](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/DiagnosticSourceUsersGuide.md) for the library, or simply taking a dependency on MEL such that a logger implementation can be provided by the user?

@johnduhart : Why we cannot use `SeriLog` ? Do we have any example of using `DiagnosticSource` in any of our libraries?

src/Medidata.MAuth.Core/MAuthCoreExtensionsV2.cs
jcarres-mdsol: Who is the original author, can @mdsol/team-51 tell us if they know why we went for extension methods here?

jcarres-mdsol: In the Ruby V2 support PR, we similarly started to review less than ideal and less than performant code and it end up in a much bigger change than expected. 
I'd say if extension methods are deemed a problem here, that you work on a PR to first make the code more flexible and once merged, you make the PR to support V2. 
It will make reviewing easier.

And I know you do not want to do many changes, because of the risk, but I think it is worse to leave the code as a weird hybrid, as it is so crucial to our operations.

hkurniawan-mdsol: I agree with this approach. Maybe we need to refactor it first to make it easier to support both versions.

prajon84: Thank you for the suggestions. I will leave this PR as is and create another PR to refactor the extensions method by creating interface. After that, I will come back to this PR. 👍

src/Medidata.MAuth.Core/Constants.cs
```diff
@@ -30,5 +30,22 @@ internal static class Constants
         public static readonly string KeyNormalizeLinesEndRegexPattern = "(?<end>-----END [A-Z ]+[-]+)$";
 
         public static readonly byte[] NewLine = Encoding.UTF8.GetBytes("\n");
+
+        public static readonly Regex AuthenticationHeaderRegexV2 = new Regex(
```
danielloganking: As this is a static Regex, it should be constructed with the `RegexOptions.Compiled` flag. This will forcibly compile the Regex at app start, trading a slower startup (and some unrecoverable memory) for more efficient evaluation: https://docs.microsoft.com/en-us/dotnet/standard/base-types/compilation-and-reuse-in-regular-expressions#compiled-regular-expressions

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -81,8 +115,14 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
                 return result;
             });
 
-        private HttpRequestMessage CreateRequest(Guid applicationUuid) =>
-            new HttpRequestMessage(HttpMethod.Get, new Uri(options.MAuthServiceUrl,
+        private HttpRequestMessage CreateRequest(Guid applicationUuid)
+        {
+            if (options.MAuthVersion.Equals(Version.V2))
```
danielloganking: Like the discussion above I want to point out that we should default to V2 and only fall back to V1 in the explicit case. As it is this makes V2 opt-in where we want it to be opt-out (as I understand the requirements).

johnduhart:
@prajon84 Pulling in Serilog is pretty heavy-weight just to solve logging, plus not all consumers may be using it. I don't think we're using DiagnosticSource anywhere, so just use `Microsoft.Extensions.Logging`'s abstractions.

src/Medidata.MAuth.Core/MAuthSigningHandler.cs
```diff
@@ -47,14 +51,20 @@ public MAuthSigningHandler(MAuthSigningOptions options, HttpMessageHandler inner
...
-        protected async override Task<HttpResponseMessage> SendAsync(
+        protected override async Task<HttpResponseMessage> SendAsync(
             HttpRequestMessage request, CancellationToken cancellationToken)
         {
             if (InnerHandler == null)
                 InnerHandler = new HttpClientHandler();
 
+            if(options.DisableV1 && options.MAuthVersion.ToString() == Version.MWS.ToString())
+                throw new InvalidVersionException
+                    ($"Signing with {options.MAuthVersion.ToString()} is no longer supported.");
```
hkurniawan-mdsol: I think the message should say
`$"Signing with {options.MAuthVersion.ToString()} is disabled."`

So it is clear that V1 is explicitly disabled in the options.

src/Medidata.MAuth.Core/Models/Version.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Runtime.Serialization;
+
+namespace Medidata.MAuth.Core.Models
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public enum Version
```
hkurniawan-mdsol: Instead of generic `Version` maybe `MAuthVersion`?

src/Medidata.MAuth.Core/UtilityExtensions.cs
```diff
@@ -61,11 +63,60 @@ public static class UtilityExtensions
...
+        /// Determines the correct MAuthHeader value 
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns>The MAuthHeader value.</returns>
+        public static string GetAuthHeaderValue(this HttpRequestMessage request)
+        {
+            //By default first check for V2
+            var authHeader = request.Headers.GetFirstValueOrDefault<string>(Constants.MAuthHeaderKeyV2)
+                             ?? request.Headers.GetFirstValueOrDefault<string>(Constants.MAuthHeaderKey);
+            if (authHeader == null)
```
hkurniawan-mdsol: Maybe use fancy null coalescing?
```cs
return authHeader ?? 
    throw new ArgumentNullException(nameof(authHeader), "The MAuth header is missing from the request.");
```

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -81,8 +95,35 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
...
+
+        //    var (uuid, payload) = authHeader.ParseAuthenticationHeader();
+
+        //    return new PayloadAuthenticationInfo()
+        //    {
+        //        ApplicationUuid = uuid,
+        //        Payload = Convert.FromBase64String(payload),
+        //        SignedTime = signedTime.FromUnixTimeSeconds()
+        //    };
+        //}
```
danielloganking: What is the purpose of this commented out code? If it isn't needed it should be removed completely.

src/Medidata.MAuth.Core/Models/Version.cs
```diff
@@ -10,12 +10,12 @@ public enum Version
...
-        [EnumMember(Value = "V1")]
-        V1 = 0,
+        [EnumMember(Value = "MWS")]
+        MWS = 0,
         /// <summary>
         /// 
         /// </summary>
-        [EnumMember(Value = "V2")]
-        V2 = 1,
+        [EnumMember(Value = "MWSV2")]
```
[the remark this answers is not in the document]
prajon84: Thanks. Will update to use Simple Enum. Attributes are not needed here.

src/Medidata.MAuth.Core/UtilityExtensions.cs
```diff
@@ -63,12 +63,60 @@ public static class UtilityExtensions
...
+            return (version == Version.MWSV2.ToString()) ?
+                Constants.MAuthTokenRequestPathV2 : Constants.MAuthTokenRequestPath;
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="version"></param>
+        /// <returns></returns>
+        public static (string mAuthHeaderKey, string mAuthTimeHeaderKey) GetHeaderKeys(this string version)
```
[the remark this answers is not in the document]
prajon84: Yes, this is being only used in test. So, I will remove it from here and add in `TestExtensions.cs` unless it is used elsewhere.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -96,12 +74,14 @@ public async Task<bool> AuthenticateRequestV2(HttpRequestMessage request)
             }
         }
 
-        private Task<ApplicationInfo> GetApplicationInfo(Guid applicationUuid) =>
+        private Task<ApplicationInfo> GetApplicationInfo(Guid applicationUuid, string version) =>
             cache.GetOrCreateAsync(applicationUuid, async entry =>
             {
+                var tokenRequestPath = version.GetMAuthTokenRequestPath();
```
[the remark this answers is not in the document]
prajon84: Thanks for the idea. Will update it.

src/Medidata.MAuth.Core/MAuthCore.cs
```diff
@@ -0,0 +1,147 @@
...
+using System.Net.Http;
+using System.Threading.Tasks;
+using Org.BouncyCastle.Crypto.Engines;
+using Org.BouncyCastle.Security;
+
+namespace Medidata.MAuth.Core
+{
+    internal class MAuthCore: IMAuthCore
+    {
+        public MAuthCore () { }
```
danielloganking: Can we remove this empty ctor as the same will be auto-generated by the compiler for easier maintainability in the future?

src/Medidata.MAuth.Core/MAuthCore.cs
```diff
@@ -0,0 +1,147 @@
...
+        /// <param name="authInfo">
+        /// The authentication information with a private key to calculate the payload with.
+        /// </param>
+        /// <returns>
+        /// A Task object which will result the request with the authentication information added when it completes.
+        /// </returns>
+        internal async Task<HttpRequestMessage> AddAuthenticationInfo(HttpRequestMessage request, PrivateKeyAuthenticationInfo authInfo)
+        {
+            var authHeader =
+                $"MWS {authInfo.ApplicationUuid.ToHyphenString()}:" +
```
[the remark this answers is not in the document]
prajon84: Yup. Might have forgot to replace at old code to use `Version` enum. There might be other places too which I need to replace this hardcoded text. Will update this and others too.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -62,12 +74,14 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
...
+                retrier = new MAuthRequestRetrier(options, (Version)Enum.Parse(typeof(Version), version));
```
danielloganking: This feels dangerous. The instance can change between V2 and V1 at runtime with opaque changes to the internals of the `request` parameter in `AuthenticateRequest` which is the public API. I feel like this further supports the idea that we should handle this logic in implementations of `IMauthCore` and try to simplify the calling code to be "version-less" as Herry suggests above.

src/Medidata.MAuth.Core/MAuthCore.cs
```diff
@@ -0,0 +1,147 @@
...
+        /// The <see cref="AuthenticationInfo"/> which holds the application uuid and the time of the signature.
+        /// </param>
+        /// <returns>A Task object which will result the SHA512 hash of the signature when it completes.</returns>
+        public async Task<byte[]> GetSignature(HttpRequestMessage request, AuthenticationInfo authInfo)
+        {
+           return new byte[][]
+                {
+                    request.Method.Method.ToBytes(), Constants.NewLine,
+                    request.RequestUri.AbsolutePath.ToBytes(), Constants.NewLine,
+                    (request.Content != null ? await request.Content.ReadAsByteArrayAsync().ConfigureAwait(false) : new byte[] { }),
```
danielloganking: I believe you can use the null coalescing operator here to simplify this line:

`await request.Content?.ReadAsByteArrayAsync().ConfigureAwait(false) ?? new byte[] {},`

src/Medidata.MAuth.Core/MAuthCoreFactory.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using Version = Medidata.MAuth.Core.Models.Version;
+
+namespace Medidata.MAuth.Core
+{
+    internal class MAuthCoreFactory
+    {
+        public static IMAuthCore Instantiate(string version = "MWS")
```
danielloganking: Why take the string here and use string comparison? This opens you up to having to deal with capitalization/culture issues and encourages poor caller hygene. Could we instead make this parameter the `Version` enum and then let the caller handle parsing the enum from a string? They are more likely to be aware of potential issues around normalization and what is appropriate and then this can be simplified to a simple `switch` with the `default` case being to throw a `NotImplementedException` (as we know that only versions specified in the enum can be passed).

If we leave it as a string I would recommend using `String.Equals` so you can specify the comparison is using the `InvariantCulture` and *is* case-sensitive (I think that's the MAuth spec right?).

version.props
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project>
   <PropertyGroup>
-    <Version>3.1.2</Version>
+    <Version>3.1.3</Version>
```
danielloganking: As best I can tell this version introduces the ability to use MAuth V2. Even though every effort is made to maintain backwards compatability, the very act of accepting a new protocol version is a breaking change. I would strongly encourage us to bump to 4.0.0

That said, this PR is already large and it might benefit to remove the V2-related code altogether and focus on the `IMauthCore` from `MauthCoreExtensions` refactor (bumping to 3.1.3) and then move to 4.0.0 with the next PR including a V2-related code.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
prajon84: Forgot to remove it. No other purpose.

version.props
johnduhart: That refactor was supposed to be completed in #46, not sure how those changed landed in this PR.

prajon84: Oh.. I have not yet updated this part yet. It was from the old refactor update. Yes I need to bump up to  `4.0.0`

danielloganking: You should rebase this branch onto the head of develop so that the changes made in #46 aren't included as part of this PR. If you need help please reach out to me and I will be happy to walk you through the branch management.

prajon84: Yes. I realized that now. seems to be off... I remember doing rebased. But, Will fix this too.

src/Medidata.MAuth.Core/UtilityExtensions.cs
```diff
@@ -63,12 +63,60 @@ public static class UtilityExtensions
...
         public static Task<bool> Authenticate(this HttpRequestMessage request, MAuthOptionsBase options) =>
-            options.MAuthVersion.Equals(Version.V2) ? new MAuthAuthenticator(options).AuthenticateRequestV2(request):
                 new MAuthAuthenticator(options).AuthenticateRequest(request);
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="authHeader"></param>
+        /// <returns></returns>
+        public static string GetVersionFromAuthenticationHeader(this string authHeader)
```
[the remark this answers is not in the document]
prajon84: Updated. https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/db4c0d11494ef55f10f5651fbdb732b04333bfec

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -280,5 +175,50 @@ private static byte[] Concat(this byte[][] values)
...
+                    $"{encodedQueryParam}&" : encodedQueryParam;
+
+                var encodedKey = Uri.EscapeUriString(key.Key); // Is this UTF8 encoding or URI encoding
+
+                var encodedValue = (queryParams[key.Key] != string.Empty)
+                    ? Uri.EscapeUriString(queryParams[key.Key])    // same here UTF8 encoding or uri encoding
+                    : "";
+                encodedQueryParam = $"{encodedQueryParam}{encodedKey}={encodedValue}";
+            }
+            return encodedQueryParam;
```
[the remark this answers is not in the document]
prajon84: Updated -> https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/db4c0d11494ef55f10f5651fbdb732b04333bfec

src/Medidata.MAuth.Core/MAuthSigningHandler.cs
prajon84: Updated -> https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/db4c0d11494ef55f10f5651fbdb732b04333bfec

src/Medidata.MAuth.Core/Models/Version.cs
prajon84: Renamed to `MAuthVersion` and updated to simple Enum without any attribute.
https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/db4c0d11494ef55f10f5651fbdb732b04333bfec

src/Medidata.MAuth.Core/UtilityExtensions.cs
prajon84: Updated -> https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/db4c0d11494ef55f10f5651fbdb732b04333bfec

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
prajon84: Updated to use Enum `version` .  https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/db4c0d11494ef55f10f5651fbdb732b04333bfec#diff-f647e3d68bb4662f9bd5f4f2474c84ecR83

src/Medidata.MAuth.Core/MAuthCoreFactory.cs
prajon84: Updated to use `MAuthVersion` enum instead of string. Also updated exception to `InvalidVersionException`.  https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/db4c0d11494ef55f10f5651fbdb732b04333bfec#diff-f8375664eed8521a12d39fce862c2b44

src/Medidata.MAuth.Core/MAuthCore.cs
prajon84: Updated  -> https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/a4dc1bcb2eb1e08387892a3c259fbba5ce0c970b#diff-543f4e6e493609bfc73aa7a79e67ec40

src/Medidata.MAuth.Core/MAuthCore.cs
prajon84: Removed -> https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/a4dc1bcb2eb1e08387892a3c259fbba5ce0c970b#diff-543f4e6e493609bfc73aa7a79e67ec40

src/Medidata.MAuth.Core/MAuthCoreV2.cs
```diff
@@ -72,23 +70,22 @@ public static bool VerifyV2(this byte[] signedData, byte[] signature, string pub
...
             var encodedAppUUid = authInfo.ApplicationUuid.ToHyphenString().ToBytes();
 
             var requestBody = request.Content != null ?
                 await request.Content.ReadAsByteArrayAsync().ConfigureAwait(false) : new byte[] { };
             // var requestBodyDigest = SHA512 hash of request body
-            // UTF-8 encode the resulting hash
             var requestBodyDigest = requestBody.AsSha512HashV2();
-            //var requestBodyDigest = Encoding.UTF8.GetString(SHA512.Create().ComputeHash(requestBody));
 
             var encodedCurrentSecondsSinceEpoch = authInfo.SignedTime.ToUnixTimeSeconds().ToString().ToBytes();
```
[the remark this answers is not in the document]
prajon84: @hkurniawan-mdsol :  If I understand your question correctly, UTF-8 encoding for `current_seconds_since_epoch ` which is same as the version V1.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -280,5 +175,50 @@ private static byte[] Concat(this byte[][] values)
 
             return result;
         }
+
+        public static IDictionary<string, string> GetQueryStringParams(this string queryString)
```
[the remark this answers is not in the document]
prajon84: Yes I have updated to use `SortedDictionary<string, string>` instead of `IDictionary<string, string>`

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -280,5 +175,50 @@ private static byte[] Concat(this byte[][] values)
...
+        {
+            var dictionary = queryString.Replace("?", "").Split('&').ToDictionary(x => x.Split('=')[0], x => x.Split('=')[1]);
```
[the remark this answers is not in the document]
prajon84: @hkurniawan-mdsol : Well, I have updated to the first way you mentioned, but I could not get anything like`new UrlEncodingParser(...)` . Else may be I can do `var query = request.RequestUri.ParseQueryString()`. What do you suggest?

src/Medidata.MAuth.Core/MAuthCoreV2.cs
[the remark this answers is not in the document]
hkurniawan-mdsol: Aah I see. the `ToBytes()` uses UTF-8 encoding to get the bytes.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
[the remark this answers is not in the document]
hkurniawan-mdsol: Sorry my bad.
I gave you the wrong package and example.

Yes, just use `var query = request.RequestUri.ParseQueryString()`

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
danielloganking: This still has mutable state which generally to me is a code smell. The fact that `retrier` is a class level variable (private field) for `MAuthAuthenticator` but changes based on the `request` parameters of the `AuthenticateRequest` method should be addressed. In this case it looks like the simplest fix would be to 1) remove the `retrier` field from the class and 2) make the `retrier` assignment in `AuthenticateRequest` into a variable assignment by adding `var`. So the new line should be:

```csharp
var retrier = new MAuthRequestRetrier(options, version);
```

src/Medidata.MAuth.Core/MAuthCore.cs
danielloganking: Safe Navigation Operator doesn't play nice with `await` so the latest code is probably the best we can do.

src/Medidata.MAuth.Core/IMAuthCore.cs
```diff
@@ -10,5 +10,9 @@ internal interface IMAuthCore
         bool Verify(byte[] signedData, byte[] signature, string publicKey);
 
         Task<byte[]> GetSignature(HttpRequestMessage request, AuthenticationInfo authInfo);
+
+        PayloadAuthenticationInfo GetAuthenticationInfo(HttpRequestMessage request);
+
+        string GetMAuthTokenRequestPath();
```
danielloganking: I don't think these methods actually needs to be a part of the public API. As it stands there are only two implementations of this interface: `MAuthCore` and `MAuthCoreV2`. Each of these is then used indirectly via `MAuthAuthenticator` to call the `GetSignature` method here (which is already part of the public interface surface) so that the signature can then be passed to the `Verify` function. As far as I can tell these are the only usages of this method so their inclusion on the core interface is likely over exposing them. I believe a more straightforward approach would be to make the following set of refactors:

1. Remove `GetAuthenticationInfo` and `GetMAuthTokenRequestPath` from the interface and make them each of private methods on their respective `IMAuthCore` implementations
2. Change the signature of `GetSignature` from `Task<byte[]> GetSignature(HttpRequestMessage request, AuthenticationInfo authInfo)` to `Task<byte[]> GetSignature(HttpRequestMessage request)`
3. Move the orchestration code to build `authInfo` from `MAuthAuthenticator.AuthenticateRequest` to the `IMAuthCore` implementations

My thinking is that `IMAuthCore` implementations should be less exposed. Ideally, only `Sign` and `Verify` would be exposed but I believe that would require perhaps too much re-structuring for this PR. But ultimately, those to me are the 2 core functions and the ones that are most likely to vary between versions. Any logic above that layer is likely to be version proof and thus encapsulating such a separation could be useful long term.

Since I'm suggesting something drastic I think it would be useful to get input from other reviewers before taking action. @hkurniawan-mdsol @johnduhart @jcarres-mdsol

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
danielloganking: I just noticed this concern also applies to the `mAuthCore` field as well.

prajon84: Updated -> https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/01e1e3ed815051d70f0c7eab699ac3b52c0035c6

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -145,5 +176,36 @@ public static byte[] Concat(this byte[][] values)
...
         }
+
+        public static SortedDictionary<string, string> SortByKeyAscending(this NameValueCollection queryStringParams)
+        {
+            var sortDictionary = new SortedDictionary<string, string>();
+            var list = queryStringParams.AllKeys.ToList();
+            list.Sort();
+            foreach (var key in list)
+            {
+                sortDictionary.Add(key, queryStringParams[key]);
```
danielloganking: I believe this will throw an exception if there are multiple query parameters with the same name. Is there any guarantee that MAuth signed calls do not repeat query parameter keys? If not how should we deal with indeterminate sorting order? If so how is that enforced (because `?q=a&q=b` is a valid query string; see [RFC1738](https://www.ietf.org/rfc/rfc1738.txt) section 3.3)?

prajon84: Removed sort as it is unnecessary. BUt, if there is params with same name then it will return this method will return as : `[0]: {[q, a,b]}`  where the param name `q` has two values `a and b`

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
prajon84: Updated for `mAuthCore` field too.

prajon84:
Can someone point me to the direction to fix this appveyor build failing?
image

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
danielloganking: This is not correct according to the MAuthV2 spec which states:

image

> Sort the parameter names by character code point in ascending order. Parameters with duplicate names should be sorted by value. For example, a parameter name that begins with an uppercase letter F precedes a parameter name that begins with a lowercase letter b.

prajon84: oops!! bummer. Does dot net have any library or extension framework that can sort in character code point ? If yes, please suggest, I will try implementing.

src/Medidata.MAuth.Core/IMAuthCore.cs
hkurniawan-mdsol: I guess it makes sense to remove `GetAuthenticationInfo` and `GetApplicationInfo` from the interface and have it as private methods in the implementing classes.

However the implementations of these methods are exactly the same. 
`GetAuthenticationInfo` is a duplication in V1 and V2 with difference in the version string.
If we move `GetApplicationInfo` then we create extra duplication.

A better way would be to move `GetAuthenticationInfo` from `IMAuthCore` to `MAuthAuthenticator` and we add methods/properties to IMAuthCore that needs different implementations for different version.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
hkurniawan-mdsol: @danielloganking I'm confused by this sentence.
>Parameters with duplicate names should be sorted by value. For example, a parameter name that begins with the uppercase letter F precedes a parameter name that begins with a lowercase letter b.

So if we encounter a duplicate name, do we sort by the value?
Why does the example talks about sorting by name again?

hkurniawan-mdsol:
@prajon84 I didn't know that we have AppVeyor build. @lschreck-mdsol do you know about this?

lschreck-mdsol:
@hkurniawan-mdsol Yes, I set up the build 😉 - it was a pilot project to see if we can use AppVeyor instead of Jenkins, but in the meantime we got Travis with cross platform .NET Core, so we did not migrate anything except this mauth library build and push.

lschreck-mdsol:
@prajon84 Try to update the XUnit packages as well as the build.ps1 script test execution (where it says `dotnet xunit` replace with `dotnet test`) and see if it fixes your issue.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
prajon84: @cmcinnes-mdsol : Since you have implemented in Ruby, can you put some light on it? I mean Can you explain with an example what does it mean?

prajon84:
> @lschreck-mdsol : Thank you. It worked for appveyor build.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
prajon84: @danielloganking @hkurniawan-mdsol  : Here as far as now I understand, if queryString  = `?key1=val1&Key1=val2` then the result should be after sort and UriEncode:  `UriEncode(Key1)=UriEncode(val2)&UriEncode(key1)=UriEncode(val1)` ?  

Since I used, `request.RequestUri.ParseQueryString()` , it puts something like this format:  `{[key1]:[val1,val2]}` which is wrong. So, I will try make update.

danielloganking: @hkurniawan-mdsol I assume the example is supposed to demonstrate sorting by codepoint (where `b`=98 and `F`=70 so `F < b`). But you are right the writing is confusing.

danielloganking: @prajon84 It seems like we need to make sure we are using `StingComparison.Ordinal` to get the proper binary/unicode ordering.

danielloganking:
BTW This is the [link to the most up to date MAuthV2 doc](https://docs.google.com/document/d/1X3X5V8LnNhKAqS53SfFJJnZe-TrjdMPw09tnO3ZpfdU/edit)

src/Medidata.MAuth.Core/IMAuthCore.cs
danielloganking: @hkurniawan-mdsol Your way definitely makes sense as I think my initial refactor would make `MAuthAuthenticator` anemic. I definitely support this.

> A better way would be to move `GetAuthenticationInfo` from `IMAuthCore` to `MAuthAuthenticator` and we add methods/properties to `IMAuthCore` that needs different implementations for different version.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
prajon84: @danielloganking : Yes I have updated now. https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/a44d7cdd813eba04e0f1d2b1ad7ebbc183106664

src/Medidata.MAuth.Core/IMAuthCore.cs
prajon84: Thanks @hkurniawan-mdsol  and @danielloganking. Will make changes.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -145,5 +176,32 @@ public static byte[] Concat(this byte[][] values)
 
             return result;
         }
+
+        /// <summary>
+        /// Builds Encoded QueryString after sort by code point and then uri encode key and values
+        /// </summary>
+        /// <param name="queryString"></param>
+        /// <returns>EncodedQueryParameter string.</returns>
+        public static string BuildEncodedQueryParams(this string queryString)
```
danielloganking: Please add unit tests for this method which align with the Ruby tests: https://github.com/mdsol/mauth-client-ruby/blob/v5.0.1/spec/signable_spec.rb#L155-L182

prajon84: Added - > https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/476d37bb45d787157ef93e11b7ae08b57a8db12b

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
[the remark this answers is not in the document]
prajon84: Updated because ParseQueryString() do not work in this case.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -145,5 +176,32 @@ public static byte[] Concat(this byte[][] values)
...
+        {
+            var encodedQueryStrings = new List<string>();
+            var queryArray = queryString.Split('&');
+            Array.Sort(queryArray, StringComparer.Ordinal);
+            Array.ForEach(queryArray, x =>
+            {
+                var keyValue = x.Split('=');
+                encodedQueryStrings.Add($"{Uri.EscapeDataString(keyValue[0])}={Uri.EscapeDataString(keyValue[1])}");
```
hkurniawan-mdsol: [nits]
How about refactor this a bit so it's easier to read.
```cs
var keyValue = x.Split('=');
var escapedKey = Uri.EscapeDataString(keyValue[0]);
var escapedValue = Uri.EscapeDataString(keyValue[1]);
encodedQueryStrings.Add($"{escapedKey}={escapedValue}");
```

prajon84: Updated -> https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/f6bcb5ef2aa99290e1e4986a5185720adab2f6c1

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -30,8 +30,8 @@ public MAuthAuthenticator(MAuthOptionsBase options)
 
             this.options = options;
 
-            _logger = LoggerFactory.Create(builder => builder.AddConsole().AddDebug())
-                .CreateLogger<MAuthAuthenticator>();
+            var factory = new LoggerFactory().AddDebug().AddConsole();
+            _logger = factory.CreateLogger<MAuthAuthenticator>();
```
johnduhart: Absolutely not. The library should not create its own logger with its own configuration. Please change to only use abstractions and for the logger to be provided by the application.

prajon84: Thank you John. I have tried to update to use logger. Please verify.
https://github.com/mdsol/mauth-client-dotnet/pull/45/commits/ebe375a0f627efbed7812d20e3bb338cebc90148

src/Medidata.MAuth.Owin/MAuthAppBuilderExtensions.cs
```diff
@@ -25,7 +28,10 @@ public static IAppBuilder UseMAuthAuthentication(this IAppBuilder app, MAuthMidd
             if (options == null)
                 throw new ArgumentNullException(nameof(options));
 
-            return app.Use<MAuthMiddleware>(options);
+            var loggerFactory = (ILoggerFactory)app.Properties.Where(x => x.Key == "ILoggerFactory")?
```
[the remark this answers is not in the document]
lschreck-mdsol: I think the logger (or in your example `ILoggerFactory`) should be passed to the middleware's constructor by the dependency injection framework and not by the user of the middleware (as you show here).

But also I think adding logging to the MAuth library would make this PR's scope too big, so I recommend to open a separate PR just for the logging part.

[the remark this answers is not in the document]
prajon84: Thanks @lschreck-mdsol . I will remove the logging part of the code from this PR. And, create another PR for logging.

prajon84:
If there are no more comments, Can this PR be approved and merged?
There will be another PR for the logging as suggested by Lazlo.

PR #46, opened 2019-07-22
[MCC-524162 Refactor to make upgrade to V2 easier

prajon84:
This PR is to refactor `MAuthCoreExtensions.cs` and implement Sign and Verification method using interface.

This PR is needed before the updates on the PR for implementing V2 logic in the PR (https://github.com/mdsol/mauth-client-dotnet/pull/45)

Please review:
@mdsol/team-51 
@johnduhart

### DIFF
--- a/src/Medidata.MAuth.Core/IMAuthCoreImplementation.cs
+++ b/src/Medidata.MAuth.Core/IMAuthCoreImplementation.cs
@@ -1,0 +1,14 @@
+﻿using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Medidata.MAuth.Core
+{
+    internal interface IMAuthCoreImplementation
+    {
+        Task<HttpRequestMessage> Sign(HttpRequestMessage request, MAuthSigningOptions options);
+
+        bool Verify(byte[] signedData, byte[] signature, string publicKey);
+
+        Task<byte[]> GetSignature(HttpRequestMessage request, AuthenticationInfo authInfo);
+    }
+}
--- a/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
+++ b/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
@@ -2,7 +2,6 @@
 using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Caching.Memory;
-using Newtonsoft.Json.Linq;
 using Org.BouncyCastle.Crypto;
 
 namespace Medidata.MAuth.Core
@@ -92,7 +91,7 @@ namespace Medidata.MAuth.Core
         /// </summary>
         /// <param name="request">The request that has the authentication information.</param>
         /// <returns>The authentication information with the payload from the request.</returns>
-        public PayloadAuthenticationInfo GetAuthenticationInfo(HttpRequestMessage request)
+        internal PayloadAuthenticationInfo GetAuthenticationInfo(HttpRequestMessage request)
         {
             var authHeader = request.Headers.GetFirstValueOrDefault<string>(Constants.MAuthHeaderKey);
 
--- a/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
+++ b/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
@@ -11,7 +11,7 @@ namespace Medidata.MAuth.Core
         private readonly MAuthOptionsBase options;
         private readonly MAuthRequestRetrier retrier;
         private readonly IMemoryCache cache = new MemoryCache(new MemoryCacheOptions());
-        private readonly MAuthCoreImplementation mAuthCore;
+        private IMAuthCore mAuthCore;
 
         public Guid ApplicationUuid => options.ApplicationUuid;
 
@@ -27,7 +27,6 @@ namespace Medidata.MAuth.Core
                 throw new ArgumentNullException(nameof(options.PrivateKey));
 
             this.options = options;
-            mAuthCore = MAuthCoreImplementation.MAuthCore;
             retrier = new MAuthRequestRetrier(options);
         }
 
@@ -35,6 +34,7 @@ namespace Medidata.MAuth.Core
         {
             try
             {
+                mAuthCore = MAuthCoreFactory.Instantiate();
                 var authInfo = GetAuthenticationInfo(request);
                 var appInfo = await GetApplicationInfo(authInfo.ApplicationUuid);
 
--- a/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
+++ b/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
@@ -2,6 +2,7 @@
 using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Caching.Memory;
+using Newtonsoft.Json.Linq;
 using Org.BouncyCastle.Crypto;
 
 namespace Medidata.MAuth.Core
@@ -11,6 +12,7 @@ namespace Medidata.MAuth.Core
         private readonly MAuthOptionsBase options;
         private readonly MAuthRequestRetrier retrier;
         private readonly IMemoryCache cache = new MemoryCache(new MemoryCacheOptions());
+        private readonly MAuthCoreImplementation mAuthCore;
 
         public Guid ApplicationUuid => options.ApplicationUuid;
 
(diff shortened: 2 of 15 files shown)


src/Medidata.MAuth.Core/MAuthCoreImplementation.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// </summary>
+        /// <param name="request">
+        /// The request which has the information (method, url and content) for the calculation.
+        /// </param>
+        /// <param name="authInfo">
+        /// The <see cref="PrivateKeyAuthenticationInfo"/> which holds the application uuid, the time of the
+        /// signature and the private key.
+        /// </param>
+        /// <returns>A task object which will result the payload as a Base64 encoded string when completed.</returns>
+        public async Task<string> CalculatePayload(HttpRequestMessage request, PrivateKeyAuthenticationInfo authInfo)
```
bvillanueva-mdsol: should this be set to internal? Not part of the interface and is not used on other classes

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -84,5 +86,32 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
         private HttpRequestMessage CreateRequest(Guid applicationUuid) =>
             new HttpRequestMessage(HttpMethod.Get, new Uri(options.MAuthServiceUrl,
                 $"{Constants.MAuthTokenRequestPath}{applicationUuid.ToHyphenString()}.json"));
+
+        /// <summary>
+        /// Extracts the authentication information from a <see cref="HttpRequestMessage"/>.
+        /// </summary>
+        /// <param name="request">The request that has the authentication information.</param>
+        /// <returns>The authentication information with the payload from the request.</returns>
+        public PayloadAuthenticationInfo GetAuthenticationInfo(HttpRequestMessage request)
```
bvillanueva-mdsol: This is not needed to be public too. Internal maybe for tests?

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -253,7 +118,7 @@ public static Stream Rewind(this Stream stream)
         public static string NormalizeLines(this string key) =>
             key.RemoveLineBreaks().InsertLineBreakAfterBegin().InsertLineBreakBeforeEnd();
 
-        private static bool IsValidPath(this string value) =>
+        public static bool IsValidPath(this string value) =>
```
bvillanueva-mdsol: revert this to private?

src/Medidata.MAuth.Core/MAuthCoreImplementation.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// Adds authentication information to a <see cref="HttpRequestMessage"/>. 
+        /// </summary>
+        /// <param name="request">The request to add the information to.</param>
+        /// <param name="authInfo">
+        /// The authentication information with a private key to calculate the payload with.
+        /// </param>
+        /// <returns>
+        /// A Task object which will result the request with the authentication information added when it completes.
+        /// </returns>
+        public async Task<HttpRequestMessage> AddAuthenticationInfo(HttpRequestMessage request, PrivateKeyAuthenticationInfo authInfo)
```
bvillanueva-mdsol: private or internal?

src/Medidata.MAuth.Core/IMAuthCoreImplementation.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Medidata.MAuth.Core
+{
+    internal interface IMAuthCoreImplementation
```
hkurniawan-mdsol: Since most of the implementation in V1 and V2 will be the same except for signature generation, I think using `abstract` class is more suitable than interface.
This is what I have in mind, sorry for the long text.

```cs
public abstract class MAuthCore
{
  public static MAuthCore Instantiate(string versionHeader)
  {
    return  versionHeader == "v1" ? new MAuthCoreV1() : new MAuthCoreV2() as MAuthCore;
  }

  public virtual bool Verify(byte[] signedData, byte[] signature, string publicKey)
  {
    return true;
  }

  public abstract Task<byte[]> GetSignature(HttpRequestMessage request, AuthenticationInfo authInfo);

  public virtual PayloadAuthenticationInfo GetAuthenticationInfo(this HttpRequestMessage request)
  {
    return null;
  }
}

public class MAuthCoreV1 : MAuthCore
{
  public override Task<byte[]> GetSignature(HttpRequestMessage request, AuthenticationInfo authInfo)
  {
    throw new NotImplementedException();
  }
}

public class MAuthCoreV2 : MAuthCore
{
  public override Task<byte[]> GetSignature(HttpRequestMessage request, AuthenticationInfo authInfo)
  {
    throw new NotImplementedException();
  }
}

// In MAuthAuthenticator.cs
internal class MAuthAuthenticator
{
  {
    try
    {
      var mauthCore = MAuthCore.Instantiate(//pass the version from request header here);
      ...
    }
    ...
  }
}
```

src/Medidata.MAuth.Core/MAuthCoreImplementation.cs
prajon84: Thank you for catching. I remember before and thought would change it but I forgot all this and others. will update to internal so that it's unit test will work too.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
prajon84: Same here will update to internal

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
prajon84: Yes.  will revert back to private as its not needed outside this class

src/Medidata.MAuth.Core/MAuthCoreImplementation.cs
prajon84: internal for the tests.

src/Medidata.MAuth.Core/IMAuthCoreImplementation.cs
johnduhart: The signature isn't the only change, there's also a change to the headers and how the signature is signed (Using RSA sign instead of public key encryption).

prajon84: Methods for Signature Creation, Signing (using RSA), Verify(Using RSA) and AddAuthenticationInfo  will have different implementation.

hkurniawan-mdsol: @johnduhart, I see.  I guess the `Interface` already extracts out the methods that need to be implemented different so this is fine.

@prajon84 Can we rename the Interface to `IMAuthCore`? Having `Implementation` in there looks weird, especially for an interface. And maybe rename the implementation class to `MAuthCore` as well.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -26,18 +27,18 @@ public MAuthAuthenticator(MAuthOptionsBase options)
                 throw new ArgumentNullException(nameof(options.PrivateKey));
 
             this.options = options;
-
+            mAuthCore = MAuthCoreImplementation.MAuthCore;
```
hkurniawan-mdsol: Two comments on this.
- We need to tell the factory method on which version of `MAuthCore` we want to instantiate. Glancing at the spec, looks like you can tell from the request header. V2 would have something like `MWSV2`. I recommend creating a factory class for instantiating `MAuthCore`. Something like
  ```cs
  public class MAuthCoreFactory
  {
      public IMAuthCore Instantiate(string version)
      {
          if (version == "MWS")
              return new MAuthCore();
          else if (version == "MWSV2")
              return new MAuthCoreV2();

          throw new Exception("Version is not recognized");
      }
  }
  ```
- If we instantiate `mauthCore` in the constructor, there's no way we can support two versions at the same time. This has to happen under
  ```cs
  {
      try 
      {
          mAuthCore = MAuthCoreFactory.Instantiate(request.Headers.GetFirstValueOrDefault<string>(Constants.MAuthHeaderKey));
          ...
      }
  }
  ```

prajon84: Thanks @hkurniawan-mdsol . I will add `MAuthCoreFactory`. But, since this PR is only refactoring, I will update the changes for V2 in other main PR. Therefore, In this case Instantiate will directly `return new MAuthCore()`

PR #47, opened 2019-08-19
[MCC-529263] Adding logging support

prajon84:
This work is the additional part which involves addition of logging support during the authentication.
The objective of logging support is to figure out how many (and who) requests are still using MWS vs MWSV2 in any given period of time.

As per the recommendation, this PR was separated from previous PR of V2.

- Added logging support for aspnetcore application by resolving  `ILoggerFactory` .
- For WebApi application, `ILoggerFactory` instance is passed along with `MAuthWebApiOptions` as optional parameter. _(This is also not found where it is being used)._
- For Owin application, currently just created `NullLoggerFactory.Instance` at the start. 

Note: Currently, `mdsol/Shadowbroker` , `mdsol/meds_extractor_jobmanager` and `mdsol/Medidata.Ampridatvir`  are only OWIN application which uses `Medidata.MAuth.Owin' and out of which two are already in the process of being converted into aspnet core application. 
Therefore, please suggest if we still need to add logging support from `Medidata.MAuth.Owin` side.

@cabbott 
@lschreck-mdsol 
@hkurniawan-mdsol 
@bvillanueva-mdsol 


src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -38,8 +41,11 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
         {
             try
             {
+                _logger.LogInformation($"Initiating Authentication of request", request);
```
hkurniawan-mdsol: I don't think that's the correct way to log the `request` data. You need a placeholder in the message like
```cs
_logger.LogInformation("Initiating Authentication of {request}", request);
```
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.2#log-message-template

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -38,8 +41,11 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
...
                 var version = request.GetAuthHeaderValue().GetVersionFromAuthenticationHeader();
 
+                _logger.LogInformation($"Authentication is for the request with {version} version.");
```
hkurniawan-mdsol: Since we don't know what kind of logging framework the hosting application will use, I assume we're not logging in `Astinus` format here?

What format does the ruby client logs in @jcarres-mdsol?

src/Medidata.MAuth.Core/UtilityExtensions.cs
```diff
@@ -65,10 +66,11 @@ public static class UtilityExtensions
         /// </summary>
         /// <param name="request">The request message to authenticate.</param>
         /// <param name="options">The MAuth options to use for the authentication.</param>
+        /// <param name="loggerFactory">The logger factory used with authentication.</param>
         /// <returns>The task for the operation that is when completes will result in <see langword="true"/> if
         /// the authentication is successful; otherwise <see langword="false"/>.</returns>
-        public static Task<bool> Authenticate(this HttpRequestMessage request, MAuthOptionsBase options) =>
-                new MAuthAuthenticator(options).AuthenticateRequest(request);
+        public static Task<bool> Authenticate(this HttpRequestMessage request, MAuthOptionsBase options, ILoggerFactory loggerFactory) =>
+                new MAuthAuthenticator(options, loggerFactory).AuthenticateRequest(request);
```
hkurniawan-mdsol: Where does this method being used?

src/Medidata.MAuth.Owin/MAuthMiddleware.cs
```diff
@@ -10,16 +11,16 @@ internal class MAuthMiddleware: OwinMiddleware
...
             this.options = options;
-            authenticator = new MAuthAuthenticator(options);
+            authenticator = new MAuthAuthenticator(options, loggerFactory);
         }
 
         public override async Task Invoke(IOwinContext context)
         {
             await context.EnsureRequestBodyStreamSeekable();
-
+            var test = context.Get<ILoggerFactory>("test");
```
hkurniawan-mdsol: This is a test/debug code right?

src/Medidata.MAuth.WebApi/MAuthAuthenticatingHandler.cs
```diff
@@ -24,11 +26,12 @@ public class MAuthAuthenticatingHandler : DelegatingHandler
...
         /// <param name="options">The options for this message handler.</param>
-        public MAuthAuthenticatingHandler(MAuthWebApiOptions options)
+        /// <param name="loggerFactory">The logger factory  used by this message handler.</param>
+        public MAuthAuthenticatingHandler(MAuthWebApiOptions options, ILoggerFactory loggerFactory = null)
         {
             this.options = options;
-
-            authenticator = new MAuthAuthenticator(options);
+            var loggerFact = loggerFactory ?? NullLoggerFactory.Instance;
+            authenticator = new MAuthAuthenticator(options, loggerFact);
```
hkurniawan-mdsol: Maybe just do 
`authenticator = new MAuthAuthenticator(options, loggerFactory ?? NullLoggerFactory.Instance);`
directly

src/Medidata.MAuth.Owin/MAuthAppBuilderExtensions.cs
```diff
@@ -24,8 +27,8 @@ public static IAppBuilder UseMAuthAuthentication(this IAppBuilder app, MAuthMidd
 
             if (options == null)
                 throw new ArgumentNullException(nameof(options));
-
-            return app.Use<MAuthMiddleware>(options);
+            var loggerFactory = NullLoggerFactory.Instance;
+            return app.Use<MAuthMiddleware>(options, loggerFactory);
```
hkurniawan-mdsol: This should be the same as https://github.com/mdsol/mauth-client-dotnet/pull/47#discussion_r315540059

src/Medidata.MAuth.Core/UtilityExtensions.cs
prajon84: This method was existing before too which calls `MAuthAuthenticator.cs`

This method was used as extension method to call Authentication.
I found this method being called from Gambit directly authenticating like this:
https://github.com/mdsol/Gambit/blob/afe161854b2a5c7b0979552303bfc5ce4fa9fa8a/Medidata.Cloud.Shared.Support/HttpModules/MAuth/MAuthHttpModule.cs#L48

src/Medidata.MAuth.Owin/MAuthMiddleware.cs
prajon84: Sorry. This was test code. Not used forgot to delete. Will remove.

src/Medidata.MAuth.AspNetCore/MAuthAppBuilderExtensions.cs
```diff
@@ -26,8 +29,9 @@ public static class MAuthAppBuilderExtensions
             if (options == null)
                 throw new ArgumentNullException(nameof(options));
 
-
-            return app.UseMiddleware<MAuthMiddleware>(options);
+            var loggerFactory = app.ApplicationServices.GetService<ILoggerFactory>() ??
+                                NullLoggerFactory.Instance;
+            return app.UseMiddleware<MAuthMiddleware>(options, loggerFactory);
```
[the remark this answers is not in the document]
prajon84: Thank you @hkurniawan-mdsol . Updated.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -52,25 +58,29 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
             }
             catch (ArgumentException ex)
             {
+                _logger.LogError($"Unable to authenticate due to invalid MAuth authentication headers. Exception: {ex.Message}");
```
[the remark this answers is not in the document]

src/Medidata.MAuth.Core/MAuthAuthenticator.cs

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
prajon84: Not sure what are you talking about here. Its just logging for information of this request is for authentication of what version `MWS` or `MWSV2`

src/Medidata.MAuth.WebApi/MAuthAuthenticatingHandler.cs
prajon84: Updated

src/Medidata.MAuth.Owin/MAuthAppBuilderExtensions.cs
prajon84: Try to update the similar way, but unable to resolve as the constructor will have three parameters. Besides, revolving around the same question: for how many applications are OWIN based that we need to support for?

hkurniawan-mdsol: I'm not sure actually. We have Ogrillon already converted (not yet released) and ShadowBroker in the process of being converted.
The only apps left that I know from T51 is Ampridatvir.
Can't speak for other teams.

Can @lschreck-mdsol and @bvillanueva-mdsol chime in as well?

src/Medidata.MAuth.Core/UtilityExtensions.cs
hkurniawan-mdsol: Okay, that's not good. 
@lschreck-mdsol any reason why we expose this method?

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
hkurniawan-mdsol: On 2nd thought, don't worry about my comment.
But I do would like to know whether
- logging is a requirement of V2?
- if it is, what kind of information we should log?

prajon84: @hkurniawan-mdsol : as per the requirement:
`mauth-client should have sufficient logging such that it will be easy to figure out how many (and who) requests are still using MWS vs MWSV2 in any given time period`

src/Medidata.MAuth.Owin/MAuthAppBuilderExtensions.cs
prajon84: @danielloganking : Do you have any suggestion here? I see 'Coder' is also Owin application.

danielloganking: Lemme look :eyes:

danielloganking: **General Feedback**
So I haven't had time to do a thorough review but I do have a general concern with the approach here, namely the use of a builder interface (`ILoggerFactory`). By using the factory and explicit passing down the dependency chain we force dependant codebases to implement their logger with respect to `ILoggerFactory` or risk having nothing logged by the MAuthMiddleware. I know, for instance, that DictionaryService only explicity provides a mechanism for resolving the `ILogger` dependency. While it is possible that `ILoggerFactory` is automagically resolvable I'm a bit worried if we rely on that exclusively.

**OWIN API Feedback**
As for the OWIN API changing, we absolutely need to include the ability to set the Logger. Adding it as a parameter on a new method overload would be the quickest recourse, but this would be hard to discover and a breaking change. However, we are bumping the major version number with this release so I assume that is fine.

prajon84: @danielloganking : Thank you for your feedback. Regarding the use of  `ILoggerFactory` , it is used based on the feedback from previous comments. 

And, for OWIN feedback:  Can you elaborate it further please?

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changes in Medidata.MAuth
 
+## v4.0.1
```
johnduhart: Did we actually release 4.0.0? Why a new version?

README.md
```diff
@@ -265,6 +265,10 @@ public static class WebApiConfig
         };
 
         config.MessageHandlers.Add(new MAuthAuthenticatingHandler(options));
+
+		// or if there is a loggerFactory in the application
+		var loggerFactory = // get the loggerFactory of the logger being used. 
+		config.MessageHandlers.Add(new MAuthAuthenticatingHandler(options, loggerFactory));
```
johnduhart: Indents.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -38,8 +41,11 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
         {
             try
             {
+                _logger.LogInformation("Initiating Authentication of {request}.", request);
```
johnduhart: How does this get serialized? There's a lot of properties in HttpRequestMessage, some of which are expensive to read.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -52,25 +58,29 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
...
             catch (RetriedRequestException ex)
             {
+                _logger.LogError(ex, "Unable to query the application information from MAuth server.");
                 throw new AuthenticationException(
                     "Could not query the application information for the application from the MAuth server.", ex);
             }
             catch (InvalidCipherTextException ex)
             {
-
+                _logger.LogError(ex, "Unable to authenticate due to invalid payload information.");
```
johnduhart: Your definition of error is different than mine.

Somebody putting garbage in their MAuth Header shouldn't result in error logs.

CHANGELOG.md
prajon84: It was merged into develop not yet released. So, same `v4.0.0` should be fine. I will update that.
Thanks.

README.md
prajon84: Updated.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
prajon84: Yeah its bad idea and not informative. Will just log only information as:
` _logger.LogInformation("Initiating Authentication of the request");`

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
prajon84: Ok. Updated to log as warning.

src/Medidata.MAuth.Owin/MAuthAppBuilderExtensions.cs
prajon84: @johnduhart : I will appreciate your feedback here too.

README.md
```diff
@@ -265,6 +265,10 @@ public static class WebApiConfig
         };
 
         config.MessageHandlers.Add(new MAuthAuthenticatingHandler(options));
+
+        // or if there is a loggerFactory in the application
+        var loggerFactory = // get the loggerFactory of the logger being used. 
+        config.MessageHandlers.Add(new MAuthAuthenticatingHandler(options, loggerFactory));
     }
```
bvillanueva-mdsol: @prajon84 Is there a sample for Asp.Net.Core mauth library in adding logger?

bvillanueva-mdsol: Or it is already picked up automatically?

src/Medidata.MAuth.Owin/MAuthMiddleware.cs
```diff
@@ -10,16 +11,15 @@ internal class MAuthMiddleware: OwinMiddleware
         private readonly MAuthMiddlewareOptions options;
         private readonly MAuthAuthenticator authenticator;
 
-        public MAuthMiddleware(OwinMiddleware next, MAuthMiddlewareOptions options): base(next)
+        public MAuthMiddleware(OwinMiddleware next, MAuthMiddlewareOptions options) : base(next)
         {
             this.options = options;
-            authenticator = new MAuthAuthenticator(options);
+            authenticator = new MAuthAuthenticator(options, NullLoggerFactory.Instance);
```
bvillanueva-mdsol: Would this work?
image

image

bvillanueva-mdsol:
Should we require logger always? All apps should have loggers in the first place. What do you guys think?

src/Medidata.MAuth.Core/UtilityExtensions.cs
lschreck-mdsol: @hkurniawan-mdsol Because the Gambit HttpModule has to use it. With MAuth we officially support only ASP.NET Core and OWIN as well as ASP.NET MVC WebAPI, but we don't support `IHttpModule` which is required by Rave EDC.

bvillanueva-mdsol:
@prajon84 Did you try these middlewares on test applications and if it logs properly? Thanks

bvillanueva-mdsol:
@prajon84 From our offline disucssion, looks like `OWIN middleware` will be not used anymore as applications are jumping to shiny dotnet core which plays well with `ILoggerFactory`. 

For OWIN middleware, I think you need to make it play well with old .net frameworks while those applications are transitioning to .net core (ex: meds_ingester, ampridatvir, etc). 

I suggest that to make owin middleware options open for any kind of logger implementation. 
You can expose the `options/config` of Owin Mauth middleware to have a `Func` to log info or log error (you can refactor it more to make it more pretty). 

Here is an example https://odetocode.com/blogs/scott/archive/2013/11/12/simple-logging-middleware-katana-part-4.aspx

You might need to provide examples for that for easy adoption. 

Challenge is how MauthAuthenticator will accept in constructor the ILoggerFactory or the custom logging function. 

Hope this will shed a light. The key is for Owin middleware implementation to be not tied up with ILoggerFactory as it is not available on previous frameworks and they might go evenutally to dotnet core if they upgrade.

prajon84:
> @bvillanueva-mdsol : Thank you for the suggestion. I will give it a shot. But, I feel we need to make another constructor for `MAuthAuthenticator.cs` to accept  `Owin.ILoggerFactory` too.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -33,13 +35,18 @@ public MAuthAuthenticator(MAuthOptionsBase options)
         /// Verifies if the <see cref="HttpRequestMessage"/> request is authenticated or not.
         /// </summary>
         /// <param name="request">The <see cref="HttpRequestMessage"/> request. </param>
+        /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> used with authentication.</param>
         /// <returns>A task object of the boolean value that verifies if the request is authenticated or not.</returns>
-        public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
+        public async Task<bool> AuthenticateRequest(HttpRequestMessage request, ILoggerFactory loggerFactory)
```
bvillanueva-mdsol: @prajon84 Current changes has duplicated core logics. Maybe you can refactor this more. 
Two public methods with same name `AuthenticateRequest` but different parameters ( like current code, one with logger, one with no logger), 
`public async Task<bool> AuthenticateRequest(HttpRequestMessage request, ILoggerFactory loggerFactory)` and
`public async Task<bool> AuthenticateRequest(HttpRequestMessage request)`

then both methods will call same one private method
`private async Task<bool> AuthenticateRequest(HttpRequestMessage request, ICustomLogger logger)` and that method will have the core implementation
```          
            try
            {
                logger.LogInformation("Initiating Authentication of the request");
                var version = request.GetAuthHeaderValue().GetVersionFromAuthenticationHeader();

                logger.LogInformation("Authentication is for {version}.", version);

                if (options.DisableV1 && version == MAuthVersion.MWS)
                    throw new InvalidVersionException($"Authentication with {version} version is disabled.");

                var authInfo = GetAuthenticationInfo(request, version);
                var appInfo = await GetApplicationInfo(authInfo.ApplicationUuid, version);

                return mAuthCore.Verify(authInfo.Payload, await mAuthCore.GetSignature(request, authInfo),
                    appInfo.PublicKey);
            }
            catch (ArgumentException ex)
            {
                logger.LogError(ex, "Unable to authenticate due to invalid MAuth authentication headers.");
                throw new AuthenticationException("The request has invalid MAuth authentication headers.", ex);
            }
            catch (RetriedRequestException ex)
            {
                logger.LogError(ex, "Unable to query the application information from MAuth server.");
                throw new AuthenticationException(
                    "Could not query the application information for the application from the MAuth server.", ex);
            }
            catch (InvalidCipherTextException ex)
            {
                logger.LogWarning(ex, "Unable to authenticate due to invalid payload information.");
                throw new AuthenticationException(
                    "The request verification failed due to an invalid payload information.", ex);
            }
            catch (InvalidVersionException ex)
            {
                logger.LogError(ex, "Unable to authenticate due to invalid version.");
                throw new InvalidVersionException(ex.Message, ex);
            }
            catch (Exception ex)
            {
                logger.LogError(ex, "Unable to authenticate due to unexpected error.");
                throw new AuthenticationException(
                    "An unexpected error occured during authentication. Please see the inner exception for details.",
                    ex
                );
            }
```

`ICustomLogger` (or whatever you will call) will be implemented with one for OWIN and other for dotnet core. 
What do you think?

prajon84: Thanks. Yes its a good idea. I have refactored and introduced  `MAuthAuthenticatorLogger.cs` class.

src/Medidata.MAuth.Core/MAuthAuthenticatorLogger.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// </summary>
+    internal class MAuthAuthenticatorLogger
+    {
+        private readonly log4net.ILog _log;
+
+        private readonly ILogger _logger;
+        public MAuthAuthenticatorLogger(ILog logger)
+        {
+            _log = logger;
+        }
```
hkurniawan-mdsol: [nits]
Instead of mixing log4net `ILog` and `ILogger` here, how about refactor it like 
```cs
// Log4NetLogger.cs

public class Log4NetLogger : ILogger
{
    private readonly ILog _log;

    public Log4NetLogger(ILog log)
    {
        _log = log;
    }

    public IDisposable BeginScope<TState>(TState state)
    {
        throw new NotImplementedException();
    }

    public bool IsEnabled(LogLevel logLevel)
    {
        switch (logLevel)
        {
            case LogLevel.Critical:
                return _log.IsFatalEnabled;
            case LogLevel.Debug:
            case LogLevel.Trace:
                return _log.IsDebugEnabled;
            case LogLevel.Error:
                return _log.IsErrorEnabled;
            case LogLevel.Information:
                return _log.IsInfoEnabled;
            case LogLevel.Warning:
                return _log.IsWarnEnabled;
            default:
                throw new ArgumentOutOfRangeException(nameof(logLevel));
        }
    }

    public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
    {
        if (!IsEnabled(logLevel))
        {
            return;
        }

        if (formatter == null)
        {
            throw new ArgumentNullException(nameof(formatter));
        }
        string message = null;
        if (null != formatter)
        {
            message = formatter(state, exception);
        }
        if (!string.IsNullOrEmpty(message) || exception != null)
        {
            switch (logLevel)
            {
                case LogLevel.Critical:
                    _log.Fatal(message);
                    break;
                case LogLevel.Debug:
                case LogLevel.Trace:
                    _log.Debug(message);
                    break;
                case LogLevel.Error:
                    _log.Error(message);
                    break;
                case LogLevel.Information:
                    _log.Info(message);
                    break;
                case LogLevel.Warning:
                    _log.Warn(message);
                    break;
                default:
                    _log.Warn($"Encountered unknown log level {logLevel}, writing out as Info.");
                    _log.Info(message, exception);
                    break;
            }
        }
    }
}
```

Then [here](https://github.com/mdsol/mauth-client-dotnet/blob/9e979f18411c4fb69171ed52b846519b2f77fd8e/src/Medidata.MAuth.Core/MAuthAuthenticator.cs#L54-L61) you refactor it to
```cs
{
    ILog log = LogManager.GetLogger(typeof(MAuthAuthenticator));
    ILogger logger;

    if (log != null)
        logger = new Log4NetLogger(log);
    else
        logger = NullLoggerFactory.Instance.CreateLogger<MAuthAuthenticator>();

    MAuthAuthenticatorLogger customLogger = new MAuthAuthenticatorLogger(logger);

    return await AuthenticateRequestCore(request, customLogger);
}
```

Please note, this is just `nits`, if this is too much work, I'm okay with current implementation.

hkurniawan-mdsol: Actually, if you implement `Log4NetLogger` above, there's probably no need for `MAuthAuthenticatorLogger` so I think this is worth trying.

prajon84: Thanks @hkurniawan-mdsol . Updated and tested too. It works.

prajon84:
If no comments, Can this be approved and merged, please ?

src/Medidata.MAuth.Core/MAuthAuthenticatorLogger.cs
bvillanueva-mdsol: Question: Why are we adding specific Logger implementation like log4net? I think Mauth client dotnet should be agnostic on the implementation. I think the `MAuthAuthenticatorLogger` was enough.

prajon84: > @bvillanueva-mdsol : Was added as per @hkurniawan-mdsol comment above.

bvillanueva-mdsol: Follow up question is what happens if the logger implemented is not logger4Net ?

bvillanueva-mdsol: Now I understand, I think because of this line on the previous commit
``` private readonly log4net.ILog _log; ```
Looks like this is assuming that Owin Mauth clients are using log4net. Is it possible if we can make this more universal?

Comment (bvillanueva-mdsol) [reply src/Medidata.MAuth.Core/MAuthAuthenticatorLogger.cs]: @prajon84 Sorry I was so late in checking this PR again. Would my previous suggestion make it more "universal" on Owin middleware, which is to expose `func` options for logging error, warning or fatal or more if needed. That way mauth-client-dotnet will be flexible enough to cater to old frameworks. New frameworks like for dotnet core are handling it very fine on your current implementation.

Comment (prajon84) [reply src/Medidata.MAuth.Core/MAuthAuthenticatorLogger.cs]: > May be I didn't understand the question correctly. But, `Log4NetLogger` is implementing `ILogger` and underneath the hood, it calls the ILogger methods.

Comment (prajon84) [reply src/Medidata.MAuth.Core/MAuthAuthenticatorLogger.cs]: @bvillanueva-mdsol : Ampridatvir, Shadowbroker, Meds-extractor_JobManager and Coder all of these applications which have OWIN are using Log4Net. Therefore I thought it should not be a problem. Besides, I did not get how to pass the Func  to the constructor of  `MAuthAuthenticator.cs`

Comment (bvillanueva-mdsol) [reply src/Medidata.MAuth.Core/MAuthAuthenticatorLogger.cs]: @prajon84 I think you are right that the `func` suggestion is not good to follow.

What I am concern more is `Medidata.MAuth.Core` is having a dependency on `log4net`
https://github.com/mdsol/mauth-client-dotnet/blob/feature/MCC-529263/src/Medidata.MAuth.Core/Medidata.MAuth.Core.csproj#L12-L14

I suggest if bringing log4net is the way, keep it on `Medidata.MAuth.Owin` project only, making an implementation of `ILogger` and make `ILogger` only as what mauth core project understands.

Personally, to make it more cleaner, I would suggest if possible is to have an extension library on top of Owin that implements ILogger as a log4net to make it extensible if there are other logger implementation to cover aside from log4net.

Comment (lschreck-mdsol) [reply src/Medidata.MAuth.Core/MAuthAuthenticatorLogger.cs]: Sorry, I completely missed this conversation and just read through this thread, but before we merge this I think at minimum these points must be satisfied:
- **No log4net dependency, or no external logging library dependency**: you must make sure that MAuth is either working with the framework provided logger interfaces (`Microsoft.Extensions.Logging`) and/or the already included library (`Microsoft.Owin.Logging`) - and I don't mean you should add Owin to the Core library...
- **No loggers passed to methods**: the method's class should receive the logger in its constructor and all the methods in the class should use that one single instance to log - if you want to log something in a "classless" extension method, either log the extension method's result, or if the extension method is too big or doing too much, you should organize it to a class instead

Comment (hkurniawan-mdsol) [reply src/Medidata.MAuth.Core/MAuthAuthenticatorLogger.cs]: Regarding this https://github.com/mdsol/mauth-client-dotnet/pull/47#discussion_r322242578

To be clear, I wasn't suggesting to hardcode dependency to `log4net`. The dependency was already there and my suggestion was more on refactoring.

But yes, we can't have a hard dependency on any logging framework.

Comment (prajon84) [reply src/Medidata.MAuth.Core/MAuthAuthenticatorLogger.cs]: Thank you @lschreck-mdsol , @hkurniawan-mdsol, @bvillanueva-mdsol .
I will remove `Log4Net` library and try to find other way to tie `Microsoft.Owin.Logging` and `Microsoft.Extensions.Logging`.
Its better late than never. :)

Comment (prajon84) [reply src/Medidata.MAuth.Core/MAuthAuthenticatorLogger.cs]: Updated. Thanks to @hkurniawan-mdsol

File: src/Medidata.MAuth.Core/MAuthAuthenticator.cs [hunk 17]
```diff
@@ -16,10 +14,11 @@ internal class MAuthAuthenticator
     {
         private readonly MAuthOptionsBase options;
         private readonly IMemoryCache cache = new MemoryCache(new MemoryCacheOptions());
+        private readonly ILogger _logger;
```
hkurniawan-mdsol: Might want to make the private variable naming consistent here.
Either prefix all privates with `_` or not at all.

src/Medidata.MAuth.WebApi/MAuthAuthenticatingHandler.cs
```diff
@@ -27,7 +29,9 @@ public class MAuthAuthenticatingHandler : DelegatingHandler
         public MAuthAuthenticatingHandler(MAuthWebApiOptions options)
         {
             this.options = options;
-            authenticator = new MAuthAuthenticator(options);
+            var loggerFactory = options.LoggerFactory ?? NullLoggerFactory.Instance;
+            var logger = loggerFactory.CreateLogger(typeof(MAuthAuthenticator));
+            authenticator = new MAuthAuthenticator(options, logger);
```
hkurniawan-mdsol: Can we refactor this bit
https://github.com/mdsol/mauth-client-dotnet/blob/73476545da4e05104c164059cf9f21777a7e022f/src/Medidata.MAuth.WebApi/MAuthAuthenticatingHandler.cs#L31-L34

to a method like `SetupHandler()` and have it called from the 2 constructors?

src/Medidata.MAuth.WebApi/MAuthAuthenticatingHandler.cs
```diff
@@ -27,7 +29,9 @@ public class MAuthAuthenticatingHandler : DelegatingHandler
         public MAuthAuthenticatingHandler(MAuthWebApiOptions options)
         {
             this.options = options;
-            authenticator = new MAuthAuthenticator(options);
+            var loggerFactory = options.LoggerFactory ?? NullLoggerFactory.Instance;
+            var logger = loggerFactory.CreateLogger(typeof(MAuthAuthenticator));
```
hkurniawan-mdsol: In the `.NET Core` and `Owin`, you have the logger name as `MAuthMiddleware` since it is created in there.
Maybe have this one the same? Instead of `MAuthAuthenticator`, use `MAuthAuthenticatingHandler`.

```cs
var logger = loggerFactory.CreateLogger<MAuthAuthenticatingHandler>();
```

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
prajon84: Updated and removed `_` to make consistent as before.

src/Medidata.MAuth.WebApi/MAuthAuthenticatingHandler.cs
prajon84: Refactored. Created  `SetupHandler()`

src/Medidata.MAuth.WebApi/MAuthAuthenticatingHandler.cs
prajon84: Updated.

src/Medidata.MAuth.WebApi/MAuthAuthenticatingHandler.cs
```diff
@@ -29,9 +29,7 @@ public class MAuthAuthenticatingHandler : DelegatingHandler
         public MAuthAuthenticatingHandler(MAuthWebApiOptions options)
         {
             this.options = options;
```
hkurniawan-mdsol: Why don't we pass `options` to `SetupHandler` as well so it can be assigned to the private `options`?

prajon84: Updated

src/Medidata.MAuth.Core/MAuthAuthenticatorLogger.cs
hkurniawan-mdsol: @lschreck-mdsol @bvillanueva-mdsol, Prajon make some changes on passing the logger (if any) to the authenticator. Please review it.

It looks good to me.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -27,19 +30,23 @@ public MAuthAuthenticator(MAuthOptionsBase options)
...
-        /// <summary>
-        /// Verifies if the <see cref="HttpRequestMessage"/> request is authenticated or not.
-        /// </summary>
-        /// <param name="request">The <see cref="HttpRequestMessage"/> request. </param>
-        /// <returns>A task object of the boolean value that verifies if the request is authenticated or not.</returns>
+        ///// <summary>
+        ///// Verifies if the <see cref="HttpRequestMessage"/> request is authenticated or not.
+        ///// </summary>
+        ///// <param name="request">The <see cref="HttpRequestMessage"/> request.</param>
+        ///// <returns>A task object of the boolean value that verifies if the request is authenticated or not.</returns>
```
lschreck-mdsol: Too many slashes.

src/Medidata.MAuth.Owin/MAuthAppBuilderExtensions.cs
```diff
@@ -25,7 +25,7 @@ public static IAppBuilder UseMAuthAuthentication(this IAppBuilder app, MAuthMidd
             if (options == null)
                 throw new ArgumentNullException(nameof(options));
 
-            return app.Use<MAuthMiddleware>(options);
+            return app.Use<MAuthMiddleware>(options, app);
```
lschreck-mdsol: You should pass the logger instance instead of the `IAppBuilder`.

src/Medidata.MAuth.WebApi/MAuthAuthenticatingHandler.cs
```diff
@@ -13,8 +15,8 @@ namespace Medidata.MAuth.WebApi
     /// </summary>
     public class MAuthAuthenticatingHandler : DelegatingHandler
     {
-        private readonly MAuthWebApiOptions options;
-        private readonly MAuthAuthenticator authenticator;
+        private MAuthWebApiOptions options;
+        private MAuthAuthenticator authenticator;
```
lschreck-mdsol: These should be still `readonly`.

src/Medidata.MAuth.WebApi/MAuthAuthenticatingHandler.cs
```diff
@@ -66,5 +65,13 @@ public MAuthAuthenticatingHandler(MAuthWebApiOptions options, HttpMessageHandler
                 .SendAsync(request, cancellationToken)
                 .ConfigureAwait(continueOnCapturedContext: false);
         }
+
+        private void SetupHandler(MAuthWebApiOptions opt)
```
lschreck-mdsol: This method is not static, why not use the private fields instead?

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
prajon84: Updated

src/Medidata.MAuth.Owin/MAuthAppBuilderExtensions.cs
prajon84: Updated

src/Medidata.MAuth.WebApi/MAuthAuthenticatingHandler.cs
prajon84: Updated

src/Medidata.MAuth.WebApi/MAuthAuthenticatingHandler.cs
prajon84: Updated

README.md
```diff
@@ -178,6 +178,16 @@ public class Startup
...
 }
 ```
+For enabling the logging from Owin application, the follofing configuration will need to be added or updated:
```
bvillanueva-mdsol: [minor comment]: `follofing` --> `following`

prajon84: Updated. Thanks.

PR #52, opened 2019-10-29
[MCC-555460] Fix deadlocks issues 

prajon84:
Recently, there had been some problem of  deadlocks on Authentication path. Therefore, this PR adds `ConfigureAwait(false)` on all of the possible cases of Authentication path to resolve those issues.
Also added `ConfigureAwaitCheck.Analyzer` to find out `await` without `ConfigureAwait(false)`

Please review:
@danielloganking 
@mdsol/team-51 

### DIFF
--- a/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
+++ b/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
@@ -52,10 +52,10 @@ namespace Medidata.MAuth.Core
 
                 var mAuthCore = MAuthCoreFactory.Instantiate(version);
                 var authInfo = GetAuthenticationInfo(request, version);
-                var appInfo = await GetApplicationInfo(authInfo.ApplicationUuid, version);
+                var appInfo = await GetApplicationInfo(authInfo.ApplicationUuid, version).ConfigureAwait(false);
 
-                return mAuthCore.Verify(authInfo.Payload, await mAuthCore.GetSignature(request, authInfo),
-                    appInfo.PublicKey);
+                return mAuthCore.Verify(authInfo.Payload, await mAuthCore.GetSignature(request, authInfo)
+                    .ConfigureAwait(false), appInfo.PublicKey);
             }
             catch (ArgumentException ex)
             {
@@ -97,9 +97,9 @@ namespace Medidata.MAuth.Core
                     applicationUuid,
                     CreateRequest, Constants.MAuthTokenRequestPath,
                     requestAttempts: (int)options.MAuthServiceRetryPolicy + 1
-                );
+                ).ConfigureAwait(false);
 
-                var result = await response.Content.FromResponse();
+                var result = await response.Content.FromResponse().ConfigureAwait(false);
 
                 entry.SetOptions(
                     new MemoryCacheEntryOptions()
--- a/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
+++ b/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
@@ -53,9 +53,9 @@ namespace Medidata.MAuth.Core
                 var mAuthCore = MAuthCoreFactory.Instantiate(version);
                 var authInfo = GetAuthenticationInfo(request, version);
                 var appInfo = await GetApplicationInfo(authInfo.ApplicationUuid, version).ConfigureAwait(false);
+                var signature = await mAuthCore.GetSignature(request, authInfo).ConfigureAwait(false);
 
-                return mAuthCore.Verify(authInfo.Payload, await mAuthCore.GetSignature(request, authInfo)
-                    .ConfigureAwait(false), appInfo.PublicKey);
+                return mAuthCore.Verify(authInfo.Payload, signature, appInfo.PublicKey);
             }
             catch (ArgumentException ex)
             {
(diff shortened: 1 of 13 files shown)


hkurniawan-mdsol:
Found 2 good articles regarding `ConfigureAwait(false)`
- https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
- https://www.reddit.com/r/dotnet/comments/7nk0uj/when_to_use_configureawaitfalse/ds2gq82/

Basically `ConfigureAwait(false)` by itself does not "prevent" deadlock. What it does it not resume continuation of the awaited task on the context of the caller. Instead, the continuation will be executed on any available thread.
The main gist is to never block on a `Task`. 
```cs
{
  ...
}

public static async Task<bool> TryAuthenticate(
  this HttpContext context, MAuthAuthenticator authenticator, bool shouldIgnoreExceptions)
{
  ...
  // Never do this
  return authenticator.AuthenticateRequest(context.Request.ToHttpRequestMessage()).Result;
}

```

johnduhart:
@hkurniawan-mdsol That's great and all, however we still have products that are not able to fully leverage async in code that requires MAuth. The library should not require continuation on the parent synccontext if it's not required.

src/Medidata.MAuth.Core/UtilityExtensions.cs
```diff
@@ -70,8 +70,8 @@ public static class UtilityExtensions
         /// <param name="logger">The logger interface used for logging.</param>
         /// <returns>The task for the operation that is when completes will result in <see langword="true"/> if
         /// the authentication is successful; otherwise <see langword="false"/>.</returns>
-        public static Task<bool> Authenticate(this HttpRequestMessage request, MAuthOptionsBase options, ILogger logger) =>
-                new MAuthAuthenticator(options, logger).AuthenticateRequest(request);
+        public static async Task<bool> Authenticate(this HttpRequestMessage request, MAuthOptionsBase options, ILogger logger) =>
+                await new MAuthAuthenticator(options, logger).AuthenticateRequest(request).ConfigureAwait(false);
```
johnduhart: This method didn't require changing as it didn't await, it was returning a Task.

danielloganking:
@hagarwal-mdsol This is so common that Microsoft is working on setting a rule in FxCop by default that enforces this: https://github.com/aspnet/Extensions/issues/1440. If possible I suggest we add the https://www.nuget.org/packages/ConfigureAwaitChecker.Analyzer/ to the project for now to ensure we don't miss any `awaits` since nothing in MAuth lib should require synchronization. We can move to FxCop or similar once MS gets around to getting it set up.

hkurniawan-mdsol:
@johnduhart agree. A library should `ConfigureAwait(false)` on task. I'm just trying to point out that `await`(asynchronous wait) is equally important.

@danielloganking nice! It would've been better if this is built into Roslyn itself.

prajon84:
Thanks guys for the feedback. Sorry was not able to respond earlier. Let me incorporate `ConfigureAwaitCheck.Analyzer` in this project.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -52,10 +52,10 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
 
                 var mAuthCore = MAuthCoreFactory.Instantiate(version);
                 var authInfo = GetAuthenticationInfo(request, version);
-                var appInfo = await GetApplicationInfo(authInfo.ApplicationUuid, version);
+                var appInfo = await GetApplicationInfo(authInfo.ApplicationUuid, version).ConfigureAwait(false);
 
-                return mAuthCore.Verify(authInfo.Payload, await mAuthCore.GetSignature(request, authInfo),
-                    appInfo.PublicKey);
+                return mAuthCore.Verify(authInfo.Payload, await mAuthCore.GetSignature(request, authInfo)
+                    .ConfigureAwait(false), appInfo.PublicKey);
```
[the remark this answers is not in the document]
hkurniawan-mdsol: @prajon84 apologies, I forgot to submit my comment above.
After you address it, I'll merge the PR.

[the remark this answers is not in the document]

src/Medidata.MAuth.Owin/MAuthOwinExtensions.cs
```diff
@@ -47,7 +47,8 @@ public static HttpRequestMessage ToHttpRequestMessage(this IOwinRequest request)
         {
             try
             {
-                return await authenticator.AuthenticateRequest(context.Request.ToHttpRequestMessage());
+                return await authenticator.AuthenticateRequest(context.Request.ToHttpRequestMessage()).
```
danielloganking: Nit:

You can return this directly without `await`.

src/Medidata.MAuth.WebApi/MAuthWebApiExtensions.cs
```diff
@@ -24,7 +24,7 @@ internal static class MAuthWebApiExtensions
         {
             try
             {
-                return await authenticator.AuthenticateRequest(request);
+                return await authenticator.AuthenticateRequest(request).ConfigureAwait(false);
```
danielloganking: Nit:

You can return this directly without `await`

src/Medidata.MAuth.Owin/MAuthOwinExtensions.cs
johnduhart: It should not be returned as it's in a `try catch`.

src/Medidata.MAuth.WebApi/MAuthWebApiExtensions.cs
johnduhart: Same, it's in a try catch

danielloganking: U right :joy:

src/Medidata.MAuth.Owin/MAuthOwinExtensions.cs
danielloganking: 🤦‍♂ u right

PR #53, opened 2019-11-02
[MCC-555926] Fixed handling of MWSV2 header throwing 403 error.

prajon84:
Recently, a bug was reported by @lgabriel-mdsol that when latest "mauth-client-dotnet" was used for Authentication, it is throwing forbidden `403` error.

**Scenario** : By default, sign on is with both `MWS` and `MWSV2` protocol. Therefore, during authentication, it checks for `MWSV2` protocol first and if it finds the header, it tries to verify `MWSV2` header.

**Problem**
During the process of verification, it has to fetch `publickey` from mauth. The mauth url is same for both protocol i.e. `<mauth_url>/mauth/v1/security_tokens/<*.json>`. And, during the fetching process, it includes both `MWS` and `MWSV2` protocol. And, this process was throwing `403` i.e. forbidden error.

**Investigation**
During this process thanks to @lgabriel-mdsol (initially) and later @cmcinnes-mdsol , we uncovered that the issue was due to when `MWSV2` header is added in the fetching process. Then, we slowly uncovered there is something wrong in the signature.

**Fix**
After finding out the fault in signature,  when we made a fix in the signature; we are successfully able to retrieve `publickey` without any error and able to verify when both `MWS` and `MWSV2` headers are added.
This PR includes the fix:

- Body digest after `SHA512` hashed needs to be converted into proper format `Hex.Encode(...)`.
- Updated `SignHash` to use `SignData` which does the hashing of the unsigned signature too. 
- Similarly, updated `VerifyHash` to use `VerifyData`.
- Updated unit tests and updated changelog for the fix input.
- Also removed unnecessary code from `MAuthAuthenticator.cs` and `MAuthRequestRetrier.cs`

Please review and merge:
@mdsol/team-51 
@lgabriel-mdsol 
@cabbott 
@cmcinnes-mdsol

### DIFF
--- a/src/Medidata.MAuth.Core/MAuthCoreV2.cs
+++ b/src/Medidata.MAuth.Core/MAuthCoreV2.cs
@@ -43,7 +43,7 @@ namespace Medidata.MAuth.Core
             var rsa = new RSACryptoServiceProvider();
             rsa.PersistKeyInCsp = false;
             rsa.ImportParameters(publicKey.AsRsaParameters());
-            return rsa.VerifyHash(signature, CryptoConfig.MapNameToOID("SHA512"), signedData);
+            return rsa.VerifyData(signature, CryptoConfig.MapNameToOID("SHA512"), signedData);
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


src/Medidata.MAuth.Core/MAuthCoreV2.cs
```diff
@@ -64,7 +64,7 @@ public async Task<byte[]> GetSignature(HttpRequestMessage request, Authenticatio
 
             var requestBody = request.Content != null ?
                 await request.Content.ReadAsByteArrayAsync().ConfigureAwait(false) : new byte[] { };
-            var requestBodyDigest = requestBody.AsSha512HashV2();
```
ykitamura-mdsol: maybe you can delete this method?

https://github.com/mdsol/mauth-client-dotnet/blob/437265fa36af510bedec1b48256989d3cbcb7e71/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs#L200-L207

prajon84: Yup true!Thanks. Deleted

prajon84:
~ ping merge please 🙏

PR #56, opened 2019-12-05
[Fix] Update aspnetcore version to 2.1 LTS 

prajon84:
Recently @awiesendangermdsol  incurred an issue and provided feedback that in updated MAuth , we have mentioned as support aspnetcore 2.0 and onwards but we had mix of references of  `Microsoft.AspNetcore.....2.2...` also.

Therefore, this PR is to address that problem and update all the aspnetcore versions to `2.1` which is LTS.

Please review:
@hkurniawan-mdsol 
@mdsol/team-51 
@awiesendangermdsol

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changes in Medidata.MAuth
 
+## v4.0.2
+- **[AspNetCore]** Update aspnetcore version to aspnetcore2.1 LTS.
+
 ## v4.0.1
 - **[Core]** Fixed default sigining with both MWS and MWSV2 instead of option selected by consuming application.
 - **[Core]** Fixed an issue related to token request path which is same for both MWS and MWSV2 protocol.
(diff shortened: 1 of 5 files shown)


hkurniawan-mdsol:
Do we need to change the target framework to `netcoreapp2.1`?
Can you please elaborate on the issue?

awiesendangermdsol:
Issue we had was when trying to add package, it failed, saying we needed reference to aspnetcore 2.2 (being currently 2.1). In any case, I would think targeting 2.1 makes more sense due to LTS?

prajon84:
> Do we need to change the target framework to `netcoreapp2.1`?
> Can you please elaborate on the issue?

@hkurniawan-mdsol : Also `netcore 2.0` is already end of support and `2.1` is soon on Dec 23, 2019. Therefore `netcore 2.1` which is LTS and most of the aspnetcore applications are atleast using or should be using `aspnetcore2.1` (if I know correctly).

hkurniawan-mdsol:
@prajon84, @awiesendangermdsol thanks for the explanation.

PR #57, opened 2019-12-06
[Refactor] Refactor log message to same as ruby, java and python

prajon84:
@ykitamura-mdsol pointed out that the log message in `mauth-client-dotnet` is not the same as other language and needs to be same in order to be searched for the logs later in sumologic.

Log message in `mauth-jvm-client`: https://github.com/mdsol/mauth-jvm-clients/blob/70d1767d744e46aa26aecbc3bafc91771c7565bd/modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java#L94

Log message in `mauth-client-ruby`: https://github.com/mdsol/mauth-client-ruby/blob/v5.0.2/lib/mauth/client/authenticator_base.rb#L51-L53

Therefore, this PR is to refactor the log message it to have the same key phrase which is: `Mauth-client attempting to authenticate request from app`

Please review:
@ykitamura-mdsol 
@mdsol/team-51

### DIFF
--- a/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
+++ b/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
@@ -45,13 +45,15 @@ namespace Medidata.MAuth.Core
                 logger.LogInformation("Initiating Authentication of the request.");
                 var version = request.GetAuthHeaderValue().GetVersionFromAuthenticationHeader();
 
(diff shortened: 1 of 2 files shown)


hkurniawan-mdsol:
I think this is fine for a few log messages however, we need to think of something more scalable when this involves a lot more messages.

@jfeltesse-mdsol if I remember correctly, you were thinking about this a while back.

PR #60, opened 2020-01-17
[MCC-575471] Fall back to V1 authentication when V2 authentication fails

prajon84:
As discussed in this [doc](https://docs.google.com/document/d/1UsKJfZeDJb3cbvi0B-z-KMXHrNEGo-GKhXi2k4OnMxU/edit#heading=h.k5ec2aaqcyc0)
and same as in ruby [mdsol/mauth-client-ruby#38](https://github.com/mdsol/mauth-client-ruby/pull/38), this PR updates the logic to fall back to V1 authentication when V2 authentication fails.

Please review:
@mdsol/team-51

### DIFF
--- a/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
+++ b/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
@@ -123,37 +123,41 @@ namespace Medidata.MAuth.Core
                 return result;
             });
 
-        private HttpRequestMessage CreateRequest(Guid applicationUuid) =>
-            new HttpRequestMessage(HttpMethod.Get, new Uri(options.MAuthServiceUrl,
-                $"{Constants.MAuthTokenRequestPath}{applicationUuid.ToHyphenString()}.json"));
-
         /// <summary>
         /// Extracts the authentication information from a <see cref="HttpRequestMessage"/>.
         /// </summary>
         /// <param name="request">The request that has the authentication information.</param>
         /// <param name="mAuthCore">Instantiation of mAuthCore class.</param>
         /// <returns>The authentication information with the payload from the request.</returns>
-        internal PayloadAuthenticationInfo GetAuthenticationInfo(HttpRequestMessage request, IMAuthCore mAuthCore)
+        internal static PayloadAuthenticationInfo GetAuthenticationInfo(HttpRequestMessage request, IMAuthCore mAuthCore)
         {
             var headerKeys = mAuthCore.GetHeaderKeys();
             var authHeader = request.Headers.GetFirstValueOrDefault<string>(headerKeys.mAuthHeaderKey);
 
             if (authHeader == null)
+            {
                 throw new ArgumentNullException(nameof(authHeader), "The MAuth header is missing from the request.");
+            }
 
             var signedTime = request.Headers.GetFirstValueOrDefault<long>(headerKeys.mAuthTimeHeaderKey);
 
             if (signedTime == default(long))
+            {
                 throw new ArgumentException("Invalid MAuth signed time header value.", nameof(signedTime));
+            }
 
             var (uuid, payload) = authHeader.ParseAuthenticationHeader();
 
-            return new PayloadAuthenticationInfo()
+            return new PayloadAuthenticationInfo
             {
                 ApplicationUuid = uuid,
                 Payload = Convert.FromBase64String(payload),
                 SignedTime = signedTime.FromUnixTimeSeconds()
             };
         }
+
+        private HttpRequestMessage CreateRequest(Guid applicationUuid) =>
+            new HttpRequestMessage(HttpMethod.Get, new Uri(options.MAuthServiceUrl,
+                $"{Constants.MAuthTokenRequestPath}{applicationUuid.ToHyphenString()}.json"));
     }
 }
--- a/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
+++ b/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
@@ -48,16 +48,17 @@ namespace Medidata.MAuth.Core
                 if (options.DisableV1 && version == MAuthVersion.MWS)
                     throw new InvalidVersionException($"Authentication with {version} version is disabled.");
 
-                var mAuthCore = MAuthCoreFactory.Instantiate(version);
-                var authInfo = GetAuthenticationInfo(request, version);
                 var logMessage = "Mauth-client attempting to authenticate request from app with mauth app uuid" +
-                        $" {authInfo.ApplicationUuid} using version {version}";
+                        $" {options.ApplicationUuid} using version {version}";
                 logger.LogInformation(logMessage);
-
-                var appInfo = await GetApplicationInfo(authInfo.ApplicationUuid, version).ConfigureAwait(false);
-                var signature = await mAuthCore.GetSignature(request, authInfo).ConfigureAwait(false);
-
-                return mAuthCore.Verify(authInfo.Payload, signature, appInfo.PublicKey);
+                var authenticated = await Authenticate(request, version).ConfigureAwait(false);
+                if (version == MAuthVersion.MWSV2 && !authenticated && !options.DisableV1)
+                {
+                    // fall back to V1 authentication
+                    authenticated = await Authenticate(request, MAuthVersion.MWS).ConfigureAwait(false);
+                    logger.LogWarning("Completed successful authentication attempt after fallback to v1");
+                }
+                return authenticated;
             }
             catch (ArgumentException ex)
             {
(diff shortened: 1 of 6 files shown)


src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -91,10 +93,20 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
...
+
+            var mAuthCore = MAuthCoreFactory.Instantiate(version);
+            var signature = await mAuthCore.GetSignature(request, authInfo).ConfigureAwait(false);
+            return mAuthCore.Verify(authInfo.Payload, signature, appInfo.PublicKey);
+        }
+
+        private Task<ApplicationInfo> GetApplicationInfo(Guid applicationUuid) =>
             cache.GetOrCreateAsync(applicationUuid, async entry =>
             {
-                var retrier = new MAuthRequestRetrier(options, version);
```
prajon84: `version` was not used in this class hence removed.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -48,16 +48,17 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
...
-                        $" {authInfo.ApplicationUuid} using version {version}";
+                        $" {options.ApplicationUuid} using version {version}";
                 logger.LogInformation(logMessage);
-
-                var appInfo = await GetApplicationInfo(authInfo.ApplicationUuid, version).ConfigureAwait(false);
-                var signature = await mAuthCore.GetSignature(request, authInfo).ConfigureAwait(false);
-
-                return mAuthCore.Verify(authInfo.Payload, signature, appInfo.PublicKey);
+                var authenticated = await Authenticate(request, version).ConfigureAwait(false);
+                if (version == MAuthVersion.MWSV2 && !authenticated && !options.DisableV1)
```
hkurniawan-mdsol: [nits]
How about change the order of the condition to
`!authenticated && version == MAuthVersion.MWSV2 && !options.DisableV1`

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -48,16 +48,17 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
...
+                {
+                    // fall back to V1 authentication
+                    authenticated = await Authenticate(request, MAuthVersion.MWS).ConfigureAwait(false);
+                    logger.LogWarning("Completed successful authentication attempt after fallback to v1");
```
hkurniawan-mdsol: [nits]
Uppercase version to `V1`?

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -91,10 +92,20 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
             }
         }
 
-        private Task<ApplicationInfo> GetApplicationInfo(Guid applicationUuid, MAuthVersion version) =>
+        private async Task<bool> Authenticate(HttpRequestMessage request, MAuthVersion version)
+        {
+            var authInfo = GetAuthenticationInfo(request, version);
```
hkurniawan-mdsol: How about instantiate `IMAuthCore` object before this line and pass it to `GetAuthenticationInfo`?
Alternatively, you can pass the `header keys` to `GetAuthenticationInfo`.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
prajon84: Was trying to match the comment that as in ruby (https://github.com/mdsol/mauth-client-ruby/blob/31d5b6255fe70f6afff93b1590a315cd374fffc4/lib/mauth/client/authenticator_base.rb#L35). But, i guess it's not necessary to be matched. Hence, updated. :)

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
prajon84: Updated to pass `IMAuthCore` instead of `version`

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
hkurniawan-mdsol: Ah, I see. Oh well either way is fine 😄

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -48,16 +48,17 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
...
                 var logMessage = "Mauth-client attempting to authenticate request from app with mauth app uuid" +
-                        $" {authInfo.ApplicationUuid} using version {version}";
+                        $" {options.ApplicationUuid} using version {version}";
                 logger.LogInformation(logMessage);
-
-                var appInfo = await GetApplicationInfo(authInfo.ApplicationUuid, version).ConfigureAwait(false);
-                var signature = await mAuthCore.GetSignature(request, authInfo).ConfigureAwait(false);
-
-                return mAuthCore.Verify(authInfo.Payload, signature, appInfo.PublicKey);
+                var authenticated = await Authenticate(request, version).ConfigureAwait(false);
```
kenyamat: I have a question. Fallback logic is placed in try block, it means that the fallback logic will not be worked if an exception is thrown in try block (e.g. throw an exception in `Authenticate`). But it looks various types of exception might be thrown, the followings are the exception list on catch block.
- ArgumentException, RetriedRequestException, InvalidCipherTextException, InvalidVersionException

Is this OK not to fallback each exception case? Please double-check.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -48,16 +48,17 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
...
+                if (!authenticated && version == MAuthVersion.MWSV2 && !options.DisableV1)
+                {
+                    // fall back to V1 authentication
```
kenyamat: How about log a information before fallback authentication like ruby-client?
https://github.com/mdsol/mauth-client-ruby/blob/a700b4e916e17c796cb54bb448a2ee7b353b5036/lib/mauth/client/authenticator_base.rb#L33

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
prajon84: @kenyamat : AFAIK, we are doing fallback to V1 only when "authenticate_v2 is false" not when some other exceptions occurs as you mentioned above. If there is some exception in V2, then it might be cause of not authenticating. It should fall back to V1 authentication only when there is no exception while V2 authentication and its simply the signature is not verified.

kenyamat: Thank you for your confirmation.

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
kenyamat: Sorry for my poor explanation. How about add logging when fallback authentication occurred like the following code, current log line is not for fallback authentication. It's enable to check occurrence of fallback authentication easily.

```
if (!authenticated && version == MAuthVersion.MWSV2 && !options.DisableV1)
{
    // fall back to V1 authentication
    logger.LogInfo("Fallback to V1 authentication."); // <-- here
    authenticated = await Authenticate(request, MAuthVersion.MWS).ConfigureAwait(false);
    logger.LogWarning("Completed successful authentication attempt after fallback to V1");
}
```

kenyamat: [This line](https://github.com/mdsol/mauth-client-ruby/blob/a700b4e916e17c796cb54bb448a2ee7b353b5036/lib/mauth/client/authenticator_base.rb#L33) in ruby code put into rescue block for fallback, I don't think it is the same meaning as [this line on .NET](https://github.com/mdsol/mauth-client-dotnet/blob/ca180f3a792d10f241fad79db105bcc2f61fbfb3/src/Medidata.MAuth.Core/MAuthAuthenticator.cs#L53)

prajon84: Ahh.. Finally I think I understood your point. Therefore, I moved `logger.LogInformation(..)` into `Authenticate(..)` method which gets logged in both cases: the initial V2 authenticate case and when it fall back to V1 authentication too. I hope it solves your question too.

kenyamat: Thank you for your understanding. I would like to keep the same processing with the other clients as possible as we can in the platform client world.

bvillanueva-mdsol:
There is a good suggestion on the `Codacy/PR Quality Review` job result. 
https://app.codacy.com/manual/thoshikawa-mdsol/mauth-client-dotnet/pullRequest?prid=4842089
@prajon84 Please assess if the suggested change is good, if good, please apply. Thanks!

prajon84:
> There is a good suggestion on the `Codacy/PR Quality Review` job result.
> https://app.codacy.com/manual/thoshikawa-mdsol/mauth-client-dotnet/pullRequest?prid=4842089
> @prajon84 Please assess if the suggested change is good, if good, please apply. Thanks!

@bvillanueva-mdsol : Well it asks to make `GetAuthenticationInfo()` method static but AFAIK, static methods should be declared inside static class and the `MAuthenticator.cs` is not static. Therefore, I need to refactor a lot for it to move to static class. I wonder how this was not a a issue before coz this code has been checked in since long time back.

prajon84:
> > There is a good suggestion on the `Codacy/PR Quality Review` job result.
> > https://app.codacy.com/manual/thoshikawa-mdsol/mauth-client-dotnet/pullRequest?prid=4842089
> > @prajon84 Please assess if the suggested change is good, if good, please apply. Thanks!
> 
> @bvillanueva-mdsol : Well it asks to make `GetAuthenticationInfo()` method static but AFAIK, static methods should be declared inside static class and the `MAuthenticator.cs` is not static. Therefore, I need to refactor a lot for it to move to static class. I wonder how this was not a a issue before coz this code has been checked in since long time back.

I updated to move the method to be static by moving to static class but now again it shows more issue as need braces `{}` around single line statement too which is unnecessary. Do we have definite set of rules that we follow or its using the basic rules?

prajon84:
~ ping merge please if not any more issues 🙏

PR #62, opened 2020-07-09
[MCC-638856] Remove double encoding for query params and normalize path for V2 signature

prajon84:
This PR implements

- Removal of "double encoding" of query string parameters and uses the same approach as that of other languages as `ruby, python and java`.  (Note: Encodes `%2B` -> `%20`  to match that of other languages.)
- Adding normalization of the `requestUri path`.
- Updated the `Changelog` and `version.props`.
- Updated the log message during  Authentication to same as that of other langs `ruby`and `python`.

Please review:
@mdsol/team-51 
@mdsol/libraries_dotnet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changes in Medidata.MAuth
 
+## v4.0.3
+ - **[Core]** Added normalization of Uri AbsolutePath.
+ - **[Core]** Added unescape step in query_string encoding to remove `double encoding`.
+
 ## v4.0.2
 - **[AspNetCore]** Update aspnetcore version to aspnetcore2.1 LTS.
 - **[Core]** Fallback to V1 protocol when V2 athentication fails.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changes in Medidata.MAuth
 
-## v4.0.3
+## v4.1.0
  - **[Core]** Added normalization of Uri AbsolutePath.
  - **[Core]** Added unescape step in query_string encoding to remove `double encoding`.
 
--- a/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
+++ b/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
@@ -235,8 +235,8 @@ namespace Medidata.MAuth.Core
                 match = match.NextMatch();
             }
 
-            // Replacing "//" and "///" into single "/"
-            normalizedPath = normalizedPath.Replace("///", "/").Replace("//", "/");
+            // Replaces multiple slashes into single "/"
+            normalizedPath = Regex.Replace(normalizedPath, "//+", "/");
 
             return normalizedPath; 
         }
--- a/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
+++ b/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
@@ -198,6 +198,7 @@ namespace Medidata.MAuth.Core
                 unEscapedQueryStrings.Add($"{unEscapedKey}={unEscapedValue}");
             });
             var unEscapedQueryArray = unEscapedQueryStrings.ToArray();
+
             Array.Sort(unEscapedQueryArray, StringComparer.Ordinal);
             Array.ForEach(unEscapedQueryArray, x =>
             {
--- a/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
+++ b/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
@@ -207,7 +207,12 @@ namespace Medidata.MAuth.Core
                 encodedQueryStrings.Add($"{escapedKey}={escapedValue}");
             });
             var result = string.Join("&", encodedQueryStrings);
-            return result.Replace("%2B","%20");
+
+            // Above encoding converts space as `%20` and `+` as `%2B`
+            // But space and `+` both needs to be converted as `%20` as per
+            // reference https://github.com/mdsol/mauth-client-ruby/blob/master/lib/mauth/request_and_response.rb#L113
+            // so this convert `%2B` into `%20` to match encodedqueryparams to that of other languages.
+            return result.Replace("%2B", "%20");
         }
 
         /// <summary>
@@ -233,8 +238,6 @@ namespace Medidata.MAuth.Core
             // Replacing "//" and "///" into single "/"
             normalizedPath = normalizedPath.Replace("///", "/").Replace("//", "/");
 
-            if (!normalizedPath.EndsWith("/") && (path.EndsWith("/") || path.EndsWith("/.") || path.EndsWith("/..")))
-                normalizedPath = $"{normalizedPath}/";
             return normalizedPath; 
         }
     }
--- a/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
+++ b/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
@@ -187,33 +187,34 @@ namespace Medidata.MAuth.Core
             if (string.IsNullOrEmpty(queryString))
                 return string.Empty;
 
-            var encodedQueryStrings = new List<string>();
-            var unEscapedQueryStrings = new List<string>();
             var queryArray = queryString.Split('&');
-            Array.ForEach(queryArray, x =>
+
+            // unescaping
+            for (int i = 0; i < queryArray.Length; i++)
             {
-                var keyValue = x.Split('=');
+                var keyValue = queryArray.ElementAt(i).Split('=');
                 var unEscapedKey = Uri.UnescapeDataString(keyValue[0]);
                 var unEscapedValue = Uri.UnescapeDataString(keyValue[1]);
-                unEscapedQueryStrings.Add($"{unEscapedKey}={unEscapedValue}");
-            });
-            var unEscapedQueryArray = unEscapedQueryStrings.ToArray();
+                queryArray[i] = queryArray[i].Replace(queryArray.ElementAt(i), $"{unEscapedKey}={unEscapedValue}");
+            }
 
-            Array.Sort(unEscapedQueryArray, StringComparer.Ordinal);
-            Array.ForEach(unEscapedQueryArray, x =>
+            // sorting
+            Array.Sort(queryArray, StringComparer.Ordinal);
+
+            // escaping
+            for (int i = 0; i < queryArray.Length; i++)
             {
-                var keyValue = x.Split('=');
+                var keyValue = queryArray.ElementAt(i).Split('=');
                 var escapedKey = Uri.EscapeDataString(keyValue[0]);
                 var escapedValue = Uri.EscapeDataString(keyValue[1]);
-                encodedQueryStrings.Add($"{escapedKey}={escapedValue}");
-            });
-            var result = string.Join("&", encodedQueryStrings);
+                queryArray[i] = queryArray[i].Replace(queryArray.ElementAt(i), $"{escapedKey}={escapedValue}");
+            }
 
             // Above encoding converts space as `%20` and `+` as `%2B`
             // But space and `+` both needs to be converted as `%20` as per
             // reference https://github.com/mdsol/mauth-client-ruby/blob/v6.0.0/lib/mauth/request_and_response.rb#L113
             // so this convert `%2B` into `%20` to match encodedqueryparams to that of other languages.
-            return result.Replace("%2B", "%20");
+            return string.Join("&", queryArray).Replace("%2B", "%20");
         }
 
         /// <summary>
@@ -227,19 +228,15 @@ namespace Medidata.MAuth.Core
                 return string.Empty;
 
             // Normalize percent encoding to uppercase i.e. %cf%80 => %CF%80
-            Regex regexHexLowcase = new Regex("%[a-f0-9]{2}", RegexOptions.Compiled);
-            var match = regexHexLowcase.Match(path);
-            string normalizedPath = path;
+            var match = new Regex("%[a-f0-9]{2}").Match(path);
             while(match.Success)
             {
-                normalizedPath = normalizedPath.Replace(match.Value, match.Value.ToUpperInvariant());
+                path = path.Replace(match.Value, match.Value.ToUpperInvariant());
                 match = match.NextMatch();
             }
 
             // Replaces multiple slashes into single "/"
-            normalizedPath = Regex.Replace(normalizedPath, "//+", "/");
-
-            return normalizedPath; 
+            return new Regex("//+").Replace(path, "/");
         }
     }
 }
--- a/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
+++ b/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
@@ -192,10 +192,10 @@ namespace Medidata.MAuth.Core
             // unescaping
             for (int i = 0; i < queryArray.Length; i++)
             {
-                var keyValue = queryArray.ElementAt(i).Split('=');
+                var keyValue = queryArray[i].Split('=');
                 var unEscapedKey = Uri.UnescapeDataString(keyValue[0]);
                 var unEscapedValue = Uri.UnescapeDataString(keyValue[1]);
-                queryArray[i] = queryArray[i].Replace(queryArray.ElementAt(i), $"{unEscapedKey}={unEscapedValue}");
+                queryArray[i] = $"{unEscapedKey}={unEscapedValue}";
             }
 
             // sorting
@@ -204,10 +204,10 @@ namespace Medidata.MAuth.Core
             // escaping
             for (int i = 0; i < queryArray.Length; i++)
             {
-                var keyValue = queryArray.ElementAt(i).Split('=');
+                var keyValue = queryArray[i].Split('=');
                 var escapedKey = Uri.EscapeDataString(keyValue[0]);
                 var escapedValue = Uri.EscapeDataString(keyValue[1]);
-                queryArray[i] = queryArray[i].Replace(queryArray.ElementAt(i), $"{escapedKey}={escapedValue}");
+                queryArray[i] = $"{escapedKey}={escapedValue}";
             }
 
             // Above encoding converts space as `%20` and `+` as `%2B`
--- a/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
+++ b/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
@@ -228,15 +228,15 @@ namespace Medidata.MAuth.Core
                 return string.Empty;
 
             // Normalize percent encoding to uppercase i.e. %cf%80 => %CF%80
-            var match = new Regex("%[a-f0-9]{2}").Match(path);
-            while(match.Success)
+            var matches = Constants.LowerCaseHexPattern.Matches(path);
+            var normalizedPath = new StringBuilder(path);
+            foreach(var item in matches)
             {
-                path = path.Replace(match.Value, match.Value.ToUpperInvariant());
-                match = match.NextMatch();
+                normalizedPath.Replace(item.ToString(), item.ToString().ToUpper());
             }
 
             // Replaces multiple slashes into single "/"
-            return new Regex("//+").Replace(path, "/");
+            return Constants.SlashPattern.Replace(normalizedPath.ToString(), "/");
         }
     }
 }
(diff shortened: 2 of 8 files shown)


src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -184,17 +184,58 @@ public static byte[] Concat(this byte[][] values)
...
+            Array.ForEach(unEscapedQueryArray, x =>
             {
                 var keyValue = x.Split('=');
                 var escapedKey = Uri.EscapeDataString(keyValue[0]);
                 var escapedValue = Uri.EscapeDataString(keyValue[1]);
                 encodedQueryStrings.Add($"{escapedKey}={escapedValue}");
             });
-            return string.Join("&", encodedQueryStrings);
+            var result = string.Join("&", encodedQueryStrings);
+            return result.Replace("%2B","%20");
```
ykitamura-mdsol, line 210: can you add a comment (and a space after `,` 😎 ) on this?

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -184,17 +184,58 @@ public static byte[] Concat(this byte[][] values)
...
+            {
+                normalizedPath = normalizedPath.Replace(match.Value, match.Value.ToUpperInvariant());
+                match = match.NextMatch();
+            }
+
+            // Replacing "//" and "///" into single "/"
+            normalizedPath = normalizedPath.Replace("///", "/").Replace("//", "/");
+
+            if (!normalizedPath.EndsWith("/") && (path.EndsWith("/") || path.EndsWith("/.") || path.EndsWith("/..")))
+                normalizedPath = $"{normalizedPath}/";
```
ykitamura-mdsol, line 237: Don't know much about dotnet 😅 but if we are using the Uri class behind the scenes, seems we don't need to do this..
https://docs.microsoft.com/en-us/dotnet/api/system.uri?view=netcore-3.1#remarks
> As part of canonicalization in the constructor for some schemes, dot-segments and empty segments (/./, /../, and //) are compacted (in other words, they are removed). The schemes for which URI will compact these sequences include http, https, tcp, net.pipe, and net.tcp. For some other schemes, these sequences are not compacted. Here's how this compacting looks in practice.

But don't trust me! 😄

prajon84: You are correct. I might have added before and later on forgot to remove.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changes in Medidata.MAuth
 
+## v4.0.3
```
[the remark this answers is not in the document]
prajon84: Updated -> [`fe32c60` (#62)](https://github.com/mdsol/mauth-client-dotnet/pull/62/commits/fe32c60b6dffacd1ed537dd26660563c0c39c100#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR3)

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
prajon84: Added -> https://github.com/mdsol/mauth-client-dotnet/pull/62/commits/fe32c60b6dffacd1ed537dd26660563c0c39c100#diff-4819a82bb7be58721857c2b409e352acR211-R215

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -184,17 +184,61 @@ public static byte[] Concat(this byte[][] values)
...
                 var escapedKey = Uri.EscapeDataString(keyValue[0]);
                 var escapedValue = Uri.EscapeDataString(keyValue[1]);
                 encodedQueryStrings.Add($"{escapedKey}={escapedValue}");
             });
-            return string.Join("&", encodedQueryStrings);
+            var result = string.Join("&", encodedQueryStrings);
+
+            // Above encoding converts space as `%20` and `+` as `%2B`
+            // But space and `+` both needs to be converted as `%20` as per
+            // reference https://github.com/mdsol/mauth-client-ruby/blob/master/lib/mauth/request_and_response.rb#L113
```
ykitamura-mdsol, line 213: master branch might change so can you use the `v6.0.0` tag?

https://github.com/mdsol/mauth-client-ruby/blob/v6.0.0/lib/mauth/request_and_response.rb#L113

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
ykitamura-mdsol: the description sounds like we can remove this also...
https://github.com/mdsol/mauth-client-dotnet/blob/fe32c60b6dffacd1ed537dd26660563c0c39c100/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs#L238-L239

prajon84: Well. Does not seem to be true for the case like this `/example///sample//` and breaks few cases
image

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
prajon84: Updated

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
ykitamura-mdsol: I see. How about using `Regex.Replace`?
```csharp
normalizedPath = Regex.Replace(normalizedPath, "//+", "/"); 
```

I know it's unlikely to happen 😅  but if you get nine "/"s (`/////////`), I think your code will return two "/"s (`//`)

prajon84: You are very right. :) Updated

ykitamura-mdsol:
looks good in a non-dotnet developer's eyes 🐥

hkurniawan-mdsol:
>looks good in a non-dotnet developer's eyes

I don't know @ykitamura-mdsol. Looking at the comment alone, looks like it came from a .NET dev. 😄

src/Medidata.MAuth.Core/MAuthCoreV2.cs
```diff
@@ -59,16 +59,19 @@ public bool Verify(byte[] signedData, byte[] signature, string publicKey)
...
                 await request.Content.ReadAsByteArrayAsync().ConfigureAwait(false) : new byte[] { };
             var requestBodyDigest = requestBody.AsSHA512Hash();
 
             var encodedCurrentSecondsSinceEpoch = authInfo.SignedTime.ToUnixTimeSeconds().ToString().ToBytes();
-            var encodedQueryParams = !string.IsNullOrEmpty(request.RequestUri.Query) ?
-                request.RequestUri.Query.Replace("?", "").BuildEncodedQueryParams().ToBytes() : new byte[] { };
+            var queryString = request.RequestUri.Query;
+            var encodedQueryParams = !string.IsNullOrEmpty(queryString) 
+                ? queryString.Substring(queryString.IndexOf("?") + 1, queryString.Length - 1)
+                    .BuildEncodedQueryParams().ToBytes()
```
jfeltesse-mdsol, line 73: [nits] IMHO the ternary would be more readable if this was on the same line as Substring but maybe the C# formatter doesn't agree?

hkurniawan-mdsol: That would go above 120 chars which is why it's being split here.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -184,17 +184,61 @@ public static byte[] Concat(this byte[][] values)
...
-            Array.Sort(queryArray, StringComparer.Ordinal);
             Array.ForEach(queryArray, x =>
+            {
+                var keyValue = x.Split('=');
+                var unEscapedKey = Uri.UnescapeDataString(keyValue[0]);
+                var unEscapedValue = Uri.UnescapeDataString(keyValue[1]);
+                unEscapedQueryStrings.Add($"{unEscapedKey}={unEscapedValue}");
+            });
+            var unEscapedQueryArray = unEscapedQueryStrings.ToArray();
+            Array.Sort(unEscapedQueryArray, StringComparer.Ordinal);
```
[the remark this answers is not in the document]
prajon84: Added new line 👍

src/Medidata.MAuth.Core/MAuthCoreV2.cs
```diff
@@ -59,16 +59,19 @@ public bool Verify(byte[] signedData, byte[] signature, string publicKey)
...
             var requestBody = request.Content != null ?
                 await request.Content.ReadAsByteArrayAsync().ConfigureAwait(false) : new byte[] { };
             var requestBodyDigest = requestBody.AsSHA512Hash();
 
             var encodedCurrentSecondsSinceEpoch = authInfo.SignedTime.ToUnixTimeSeconds().ToString().ToBytes();
-            var encodedQueryParams = !string.IsNullOrEmpty(request.RequestUri.Query) ?
-                request.RequestUri.Query.Replace("?", "").BuildEncodedQueryParams().ToBytes() : new byte[] { };
+            var queryString = request.RequestUri.Query;
+            var encodedQueryParams = !string.IsNullOrEmpty(queryString) 
+                ? queryString.Substring(queryString.IndexOf("?") + 1, queryString.Length - 1)
```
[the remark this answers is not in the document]
prajon84: Yup. You are correct. Updated ->https://github.com/mdsol/mauth-client-dotnet/pull/62/commits/a16d0e0bc4d2be4f33b4fa098361114deaba8c37#diff-941a5f0826d6809e1b9eb07dbe49ae84R72

src/Medidata.MAuth.Core/MAuthCoreV2.cs
prajon84: After the `Substring(1)` everything fitted in one line i.e. within 120 chars. :)

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -184,17 +184,61 @@ public static byte[] Concat(this byte[][] values)
...
+
             var encodedQueryStrings = new List<string>();
+            var unEscapedQueryStrings = new List<string>();
             var queryArray = queryString.Split('&');
-            Array.Sort(queryArray, StringComparer.Ordinal);
             Array.ForEach(queryArray, x =>
+            {
+                var keyValue = x.Split('=');
+                var unEscapedKey = Uri.UnescapeDataString(keyValue[0]);
+                var unEscapedValue = Uri.UnescapeDataString(keyValue[1]);
```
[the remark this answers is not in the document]
prajon84: The reason for unescaping is : because before we were doing double encoding the query params because of which every other language was creating different "signature_to_sign". Therefore, we need to unescape the query string we receive before we process. So, the steps are : 1. Unescape 2. Sort 3. Escape (all the necessary characters.)

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -184,17 +184,62 @@ public static byte[] Concat(this byte[][] values)
...
+        /// </summary>
+        /// <param name="path"></param>
+        /// <returns>Normalized Uri Resource Path</returns>
+        public static string NormalizeUriPath(this string path)
+        {
+            if (string.IsNullOrEmpty(path))
+                return string.Empty;
+
+            // Normalize percent encoding to uppercase i.e. %cf%80 => %CF%80
+            Regex regexHexLowcase = new Regex("%[a-f0-9]{2}", RegexOptions.Compiled);
```
johnduhart, line 230: Marking this as `Compiled` does nothing because this instance is not persisted.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -184,17 +184,62 @@ public static byte[] Concat(this byte[][] values)
...
+            var match = regexHexLowcase.Match(path);
+            string normalizedPath = path;
+            while(match.Success)
+            {
+                normalizedPath = normalizedPath.Replace(match.Value, match.Value.ToUpperInvariant());
+                match = match.NextMatch();
+            }
+
+            // Replaces multiple slashes into single "/"
+            normalizedPath = Regex.Replace(normalizedPath, "//+", "/");
```
johnduhart, line 240: Non-compiled regex on a hot path.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -184,17 +184,62 @@ public static byte[] Concat(this byte[][] values)
...
+            // so this convert `%2B` into `%20` to match encodedqueryparams to that of other languages.
+            return result.Replace("%2B", "%20");
+        }
+
+        /// <summary>
+        /// Normalizes the UriPath
+        /// </summary>
+        /// <param name="path"></param>
+        /// <returns>Normalized Uri Resource Path</returns>
+        public static string NormalizeUriPath(this string path)
```
johnduhart, line 224: I'm concerned about the performance of this method, it's repeatedly re-allocating a string each time it makes a replacement to the path. Can we not optimize this?

tests/Medidata.MAuth.Tests/MAuthCoreExtensionsTests.cs
```diff
@@ -35,5 +36,62 @@ public static void BuildEncodedQueryParams_WillHandlesQueryStringWithEmptyValues
...
+        {
+            string path =null;
+            Assert.Empty(path.NormalizeUriPath());
+        }
+
+        [Fact]
+        public static void NormalizeUriPath_WithValues()
+        {
+            var testcases = CreateUriPathArray();
+            for(int i = 0; i <= testcases.GetUpperBound(0); i++)
```
johnduhart, line 74: Please use theories for this. https://andrewlock.net/creating-parameterised-tests-in-xunit-with-inlinedata-classdata-and-memberdata/

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -184,17 +184,62 @@ public static byte[] Concat(this byte[][] values)
         /// <returns>EncodedQueryParameter string.</returns>
         public static string BuildEncodedQueryParams(this string queryString)
         {
+            if (string.IsNullOrEmpty(queryString))
+                return string.Empty;
+
             var encodedQueryStrings = new List<string>();
+            var unEscapedQueryStrings = new List<string>();
             var queryArray = queryString.Split('&');
```
johnduhart, line 192: This is wasteful. We don't need a variable-size array here, we know how many elements will be in this list based on the length of `queryArray`.

Given we don't need to refer back to the original, can we please just do an inplace update of `queryArray` with unescaped, sorted, and then escaped values?

prajon84:
> Performance concerns. Please ping @mdsol/libraries_dotnet for changes to core .NET libraries.
Thanks. I will remember to do it. :)

prajon84:
> Additionally, are the edge cases we're trying to cover part of [mdsol/mauth-protocol-test-suite](https://github.com/mdsol/mauth-protocol-test-suite)? Is there a plan to add support?

Yes, there are two more stories on the board that needs to be tackled. https://jira.mdsol.com/browse/MCC-638046  and https://jira.mdsol.com/browse/MCC-575820

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
prajon84: So, `Compiled` is not necessary here?

tests/Medidata.MAuth.Tests/MAuthCoreExtensionsTests.cs
prajon84: Updated -> https://github.com/mdsol/mauth-client-dotnet/pull/62/commits/2a6e7fef2f8201f4cc6c08cb00c48ec48e870ce9

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
prajon84: Updated as this (removed unnecessary creation of variables) -> https://github.com/mdsol/mauth-client-dotnet/pull/62/commits/a3e46f8b68d86115e3e6689ac02ad83c43d4aeb0

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
prajon84: Updated -> https://github.com/mdsol/mauth-client-dotnet/pull/62/commits/a3e46f8b68d86115e3e6689ac02ad83c43d4aeb0#diff-4819a82bb7be58721857c2b409e352acR239

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
prajon84: Updated to use same string `path` and update its value and return -> https://github.com/mdsol/mauth-client-dotnet/pull/62/commits/a3e46f8b68d86115e3e6689ac02ad83c43d4aeb0#diff-4819a82bb7be58721857c2b409e352acL224-R239

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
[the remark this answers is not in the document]
hkurniawan-mdsol: I believe URI inside a .NET web app is already escaped, which is why we don't even need to do anything.
Unless we need to sort the query string keys, then we have to unescape, sort, and escape it again.

Can you adhoc-ing by sending a request and put a breakpoint on this line?
https://github.com/mdsol/mauth-client-dotnet/blob/3d0a5fb804b6f3bc84b2f2c51348542ad95b2e6d/src/Medidata.MAuth.Core/MAuthCoreV2.cs#L61
Check to see if query string is already escaped or not?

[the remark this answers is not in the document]
prajon84: Yes, which you said is right. But we need to do this because:
1. Need to sort the query string parameters before they are escaped.
2. Query string parameters already escaped by .net web app  but not all params are escaped as its required. Therefore, it makes more confusion and had to individually replace those. 

That is why, unescape, sort and escape again process is followed.(in all languages) to create the correct and same signature.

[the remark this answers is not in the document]
ykitamura-mdsol: > I believe URI inside a .NET web app is already escaped

This is same in ruby and other languages but encoding for some characters (e.g. space, tilde) are varied among languages or even versions of languages:
https://github.com/mdsol/mauth-client-ruby/blob/v6.0.0/lib/mauth/request_and_response.rb#L110 

We've discussed with T16 and decided to do "Unescape then Encode" in all mauth-clients to generate standardized mAuth signatures.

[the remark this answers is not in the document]
hkurniawan-mdsol: Got it! So the `escaped` done by the framework produces different characters to other languages.

Cool! Thanks guys.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
johnduhart:
``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.17763.1217 (1809/October2018Update/Redstone5)
Intel Core i7-7820HQ CPU 2.90GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.301
  [Host] : .NET Core 3.1.5 (CoreCLR 4.700.20.26901, CoreFX 4.700.20.27001), X64 RyuJIT


```
|   Method |         Mean |      Error |     StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|--------- |-------------:|-----------:|-----------:|-------:|-------:|------:|----------:|
| Original | 1,522.822 μs | 29.9858 μs | 64.5476 μs | 3.9063 | 1.9531 |     - |   16.2 KB |
|  Updated |     3.785 μs |  0.0753 μs |  0.1971 μs | 0.5951 |      - |     - |   2.44 KB |


That's given a test case of of `"/%2a%80/%2a%cf///example/test%20path/info"`, I think that's okay for now.

prajon84:
@hkurniawan-mdsol or @lschreck-mdsol : Can you please trigger the appveyor build for this branch? 
And, if one the approval and can merge too please 🙏

PR #64, opened 2020-07-27
[MCC-575820] Incorporate protocol test suite and implement

prajon84:
Following the changes made in https://github.com/mdsol/mauth-client-python/pull/15
and https://github.com/mdsol/mauth-client-ruby/pull/39

Updates includes

- Add [mauth-protocol-test-suite](https://github.com/mdsol/mauth-protocol-test-suite)
- Updated Sorting part in `BuildEncodedQueryParams` 

Please review 
@mdsol/libraries_dotnet 
@mdsol/team-51

### DIFF
--- a/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
+++ b/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
@@ -188,33 +188,27 @@ namespace Medidata.MAuth.Core
                 return string.Empty;
 
             var queryArray = queryString.Split('&');
+            var unescapedKeysAndValues = new KeyValuePair<string, string>[queryArray.Count()];
 
             // unescaping
             for (int i = 0; i < queryArray.Length; i++)
             {
                 var keyValue = queryArray[i].Split('=');
-                var unEscapedKey = Uri.UnescapeDataString(keyValue[0]);
-                var unEscapedValue = Uri.UnescapeDataString(keyValue[1]);
-                queryArray[i] = $"{unEscapedKey}={unEscapedValue}";
+                unescapedKeysAndValues[i] = new KeyValuePair<string, string>(
+                    Uri.UnescapeDataString(keyValue[0]), Uri.UnescapeDataString(keyValue[1]));
             }
 
-            // sorting
-            Array.Sort(queryArray, StringComparer.Ordinal);
-
-            // escaping
-            for (int i = 0; i < queryArray.Length; i++)
-            {
-                var keyValue = queryArray[i].Split('=');
-                var escapedKey = Uri.EscapeDataString(keyValue[0]);
-                var escapedValue = Uri.EscapeDataString(keyValue[1]);
-                queryArray[i] = $"{escapedKey}={escapedValue}";
-            }
+            // sorting and escaping
+            var escapedKeyValues = unescapedKeysAndValues
+                .OrderBy(kv => kv.Key, StringComparer.Ordinal)
+                .ThenBy(kv => kv.Value, StringComparer.Ordinal)
+                .Select(kv => $"{Uri.EscapeDataString(kv.Key)}={Uri.EscapeDataString(kv.Value)}");
 
             // Above encoding converts space as `%20` and `+` as `%2B`
             // But space and `+` both needs to be converted as `%20` as per
             // reference https://github.com/mdsol/mauth-client-ruby/blob/v6.0.0/lib/mauth/request_and_response.rb#L113
             // so this convert `%2B` into `%20` to match encodedqueryparams to that of other languages.
-            return string.Join("&", queryArray).Replace("%2B", "%20");
+            return string.Join("&", escapedKeyValues).Replace("%2B", "%20");
         }
 
         /// <summary>
--- a/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
+++ b/src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
@@ -188,7 +188,7 @@ namespace Medidata.MAuth.Core
                 return string.Empty;
 
             var queryArray = queryString.Split('&');
-            var unescapedKeysAndValues = new KeyValuePair<string, string>[queryArray.Count()];
+            var unescapedKeysAndValues = new KeyValuePair<string, string>[queryArray.Length];
 
             // unescaping
             for (int i = 0; i < queryArray.Length; i++)
--- a/tests/Medidata.MAuth.Tests/Infrastructure/MAuthServerHandler.cs
+++ b/tests/Medidata.MAuth.Tests/Infrastructure/MAuthServerHandler.cs
@@ -62,8 +62,7 @@ namespace Medidata.MAuth.Tests.Infrastructure
                             Uuid = isProtocolSuiteTest ? SigningAppUuid : clientUuid,
                             Name = "Medidata.MAuth.Tests",
                             CreationDate = new DateTimeOffset(2016, 8, 1, 0, 0, 0, TimeSpan.Zero),
-                            PublicKey = isProtocolSuiteTest 
-                            ? SigningPublicKey : TestExtensions.ClientPublicKey
+                            PublicKey = isProtocolSuiteTest ? SigningPublicKey : TestExtensions.ClientPublicKey
                         }
                     })
                 )
--- a/tests/Medidata.MAuth.Tests/Infrastructure/MAuthServerHandler.cs
+++ b/tests/Medidata.MAuth.Tests/Infrastructure/MAuthServerHandler.cs
@@ -11,24 +11,30 @@ namespace Medidata.MAuth.Tests.Infrastructure
 {
     internal class MAuthServerHandler : HttpMessageHandler
     {
-        private static readonly Guid clientUuid = new Guid("192cce84-8466-490e-b03e-074f82da3ee2");
-        private int currentNumberOfAttempts = 0;
+        private static readonly Guid _clientUuid = new Guid("192cce84-8466-490e-b03e-074f82da3ee2");
+        private int _currentNumberOfAttempts = 0;
 
         public int SucceedAfterThisManyAttempts { get; set; } = 1;
 
-        private static ProtocolTestSuiteHelper _protocolSuiteHelper = new ProtocolTestSuiteHelper();
+        private static string _signingPublicKey;
+        private static Guid _signingAppUuid;
 
-        public static readonly string SigningPublicKey = _protocolSuiteHelper.GetPublicKey().Result;
-        public static readonly Guid SigningAppUuid = _protocolSuiteHelper.ReadSignInAppUuid().Result;
+        public async Task<MAuthServerHandler> InitializeAsync()
+        {
+            var protocolSuite = new ProtocolTestSuiteHelper();
+            _signingPublicKey = await protocolSuite.GetPublicKey();
+            _signingAppUuid = await protocolSuite.ReadSignInAppUuid();
+            return this;
+        }
 
         protected override async Task<HttpResponseMessage> SendAsync(
             HttpRequestMessage request, CancellationToken cancellationToken)
         {
-            currentNumberOfAttempts += 1;
+            _currentNumberOfAttempts += 1;
             var version = request.GetAuthHeaderValue().GetVersionFromAuthenticationHeader();
             var mAuthCore = MAuthCoreFactory.Instantiate(version);
 
-            if (currentNumberOfAttempts < SucceedAfterThisManyAttempts)
+            if (_currentNumberOfAttempts < SucceedAfterThisManyAttempts)
                 return new HttpResponseMessage(HttpStatusCode.ServiceUnavailable);
             var authInfo = MAuthAuthenticator.GetAuthenticationInfo(request, mAuthCore);
 
@@ -39,17 +45,17 @@ namespace Medidata.MAuth.Tests.Infrastructure
                 return new HttpResponseMessage(HttpStatusCode.Unauthorized) { RequestMessage = request };
 
             if (!request.RequestUri.AbsolutePath.Equals(
-                $"{Constants.MAuthTokenRequestPath}{clientUuid.ToHyphenString()}.json",
+                $"{Constants.MAuthTokenRequestPath}{_clientUuid.ToHyphenString()}.json",
                 StringComparison.OrdinalIgnoreCase) &&
                 !request.RequestUri.AbsolutePath.Equals(
-                $"{Constants.MAuthTokenRequestPath}{SigningAppUuid.ToHyphenString()}.json",
+                $"{Constants.MAuthTokenRequestPath}{_signingAppUuid.ToHyphenString()}.json",
                 StringComparison.OrdinalIgnoreCase))
             {
                 return new HttpResponseMessage(HttpStatusCode.NotFound) { RequestMessage = request };
             }
 
             bool isProtocolSuiteTest = request.RequestUri.AbsolutePath.Contains(
-                SigningAppUuid.ToHyphenString());
+                _signingAppUuid.ToHyphenString());
 
             return new HttpResponseMessage(HttpStatusCode.OK)
             {
@@ -59,10 +65,10 @@ namespace Medidata.MAuth.Tests.Infrastructure
                     {
                         security_token = new ApplicationInfo()
                         {
-                            Uuid = isProtocolSuiteTest ? SigningAppUuid : clientUuid,
+                            Uuid = isProtocolSuiteTest ? _signingAppUuid : _clientUuid,
                             Name = "Medidata.MAuth.Tests",
                             CreationDate = new DateTimeOffset(2016, 8, 1, 0, 0, 0, TimeSpan.Zero),
-                            PublicKey = isProtocolSuiteTest ? SigningPublicKey : TestExtensions.ClientPublicKey
+                            PublicKey = isProtocolSuiteTest ? _signingPublicKey : TestExtensions.ClientPublicKey
                         }
                     })
                 )
--- a/tests/Medidata.MAuth.Tests/Infrastructure/MAuthServerHandler.cs
+++ b/tests/Medidata.MAuth.Tests/Infrastructure/MAuthServerHandler.cs
@@ -4,7 +4,7 @@ using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 using Medidata.MAuth.Core;
-using Microsoft.Extensions.Logging.Abstractions;
+using Medidata.MAuth.Tests.ProtocolTestSuite;
 using Newtonsoft.Json;
 
 namespace Medidata.MAuth.Tests.Infrastructure
@@ -15,6 +15,11 @@ namespace Medidata.MAuth.Tests.Infrastructure
         private int currentNumberOfAttempts = 0;
 
         public int SucceedAfterThisManyAttempts { get; set; } = 1;
+
+        private static ProtocolTestSuiteHelper _protocolSuiteHelper = new ProtocolTestSuiteHelper();
+
+        public static readonly string SigningPublicKey = _protocolSuiteHelper.GetPublicKey().Result;
+        public static readonly Guid SigningAppUuid = _protocolSuiteHelper.ReadSignInAppUuid().Result;
 
         protected override async Task<HttpResponseMessage> SendAsync(
             HttpRequestMessage request, CancellationToken cancellationToken)
@@ -35,8 +40,16 @@ namespace Medidata.MAuth.Tests.Infrastructure
 
             if (!request.RequestUri.AbsolutePath.Equals(
                 $"{Constants.MAuthTokenRequestPath}{clientUuid.ToHyphenString()}.json",
+                StringComparison.OrdinalIgnoreCase) &&
+                !request.RequestUri.AbsolutePath.Equals(
+                $"{Constants.MAuthTokenRequestPath}{SigningAppUuid.ToHyphenString()}.json",
                 StringComparison.OrdinalIgnoreCase))
+            {
                 return new HttpResponseMessage(HttpStatusCode.NotFound) { RequestMessage = request };
+            }
+
+            bool isProtocolSuiteTest = request.RequestUri.AbsolutePath.Contains(
+                SigningAppUuid.ToHyphenString());
 
             return new HttpResponseMessage(HttpStatusCode.OK)
             {
@@ -46,10 +59,11 @@ namespace Medidata.MAuth.Tests.Infrastructure
                     {
                         security_token = new ApplicationInfo()
                         {
-                            Uuid = clientUuid,
+                            Uuid = isProtocolSuiteTest ? SigningAppUuid : clientUuid,
                             Name = "Medidata.MAuth.Tests",
                             CreationDate = new DateTimeOffset(2016, 8, 1, 0, 0, 0, TimeSpan.Zero),
-                            PublicKey = TestExtensions.ClientPublicKey
+                            PublicKey = isProtocolSuiteTest 
+                            ? SigningPublicKey : TestExtensions.ClientPublicKey
                         }
                     })
                 )
--- a/tests/Medidata.MAuth.Tests/Infrastructure/MAuthServerHandler.cs
+++ b/tests/Medidata.MAuth.Tests/Infrastructure/MAuthServerHandler.cs
@@ -11,6 +11,8 @@ namespace Medidata.MAuth.Tests.Infrastructure
 {
     internal class MAuthServerHandler : HttpMessageHandler
     {
+        private MAuthServerHandler() { }
+
         private static readonly Guid _clientUuid = new Guid("192cce84-8466-490e-b03e-074f82da3ee2");
         private int _currentNumberOfAttempts = 0;
 
@@ -19,12 +21,18 @@ namespace Medidata.MAuth.Tests.Infrastructure
         private static string _signingPublicKey;
         private static Guid _signingAppUuid;
 
-        public async Task<MAuthServerHandler> InitializeAsync()
+        private async Task<MAuthServerHandler> InitializeAsync()
         {
             var protocolSuite = new ProtocolTestSuiteHelper();
             _signingPublicKey = await protocolSuite.GetPublicKey();
             _signingAppUuid = await protocolSuite.ReadSignInAppUuid();
             return this;
+        }
+
+        public static Task<MAuthServerHandler> CreateAsync()
+        {
+            var ret = new MAuthServerHandler();
+            return ret.InitializeAsync();
         }
 
         protected override async Task<HttpResponseMessage> SendAsync(
--- a/tests/Medidata.MAuth.Tests/Infrastructure/TestExtensions.cs
+++ b/tests/Medidata.MAuth.Tests/Infrastructure/TestExtensions.cs
@@ -20,13 +20,13 @@ namespace Medidata.MAuth.Tests.Infrastructure
         public static readonly string ServerPrivateKey = GetKeyFromResource(nameof(ServerPrivateKey)).Result;
         public static readonly string ServerPublicKey = GetKeyFromResource(nameof(ServerPublicKey)).Result;
 
-        public static MAuthOptionsBase ServerOptions => new MAuthTestOptions()
+        public static MAuthOptionsBase ServerOptions(MAuthServerHandler serverHandler) => new MAuthTestOptions()
         {
             ApplicationUuid = ServerUuid,
             MAuthServiceUrl = TestUri,
             PrivateKey = ServerPrivateKey,
             MAuthServiceRetryPolicy = MAuthServiceRetryPolicy.RetryOnce,
-            MAuthServerHandler = new MAuthServerHandler()
+            MAuthServerHandler = serverHandler
         };
 
         public static MAuthSigningOptions ClientOptions(DateTimeOffset signedTime) => new MAuthSigningOptions()
@@ -37,18 +37,18 @@ namespace Medidata.MAuth.Tests.Infrastructure
         };
 
         public static MAuthOptionsBase GetServerOptionsWithAttempts(MAuthServiceRetryPolicy policy,
-            bool shouldSucceedWithin) =>
-            new MAuthTestOptions()
+            bool shouldSucceedWithin, MAuthServerHandler serverHandler)
+        {
+            serverHandler.SucceedAfterThisManyAttempts = (int)policy + (shouldSucceedWithin ? 1 : 2);
+            return new MAuthTestOptions()
             {
                 ApplicationUuid = ServerUuid,
                 MAuthServiceUrl = TestUri,
                 PrivateKey = ServerPrivateKey,
                 MAuthServiceRetryPolicy = policy,
-                MAuthServerHandler = new MAuthServerHandler()
-                {
-                    SucceedAfterThisManyAttempts = (int)policy + (shouldSucceedWithin ? 1 : 2)
-                }
+                MAuthServerHandler = serverHandler
             };
+        }
 
         public static Task<string> GetStringFromResource(string resourceName)
         {
--- a/tests/Medidata.MAuth.Tests/Infrastructure/TestExtensions.cs
+++ b/tests/Medidata.MAuth.Tests/Infrastructure/TestExtensions.cs
@@ -7,7 +7,6 @@ using System.Threading.Tasks;
 using Medidata.MAuth.Core;
 using Newtonsoft.Json;
 using Medidata.MAuth.Core.Models;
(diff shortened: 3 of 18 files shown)


tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
```diff
@@ -0,0 +1,131 @@
...
+            var currentDirectory = Environment.CurrentDirectory;
+            _testSuitePath = Environment.GetEnvironmentVariable("TEST_SUITE_PATH") != null
+                ? Environment.GetEnvironmentVariable("TEST_SUITE_PATH")
+                : Path.GetFullPath(Path.Combine(currentDirectory, @"..\..\..\..\..\..\mauth-protocol-test-suite"));
+            _testCasePath = $"{_testSuitePath}/protocols/MWSV2";
+        }
+
+        public async Task<SigningConfig> LoadSigningConfig()
+        {
+            var configFile = $"{_testSuitePath}/signing-config.json";
```
johnduhart, line 26: Please use `Path.Join` for concating paths.

tests/Medidata.MAuth.Tests/MAuthProtocolSuiteTests.cs
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        [Theory, MemberData(nameof(TestCases))]
+        public async Task MAuth_Execute_ProtocolTestSuite(string caseName)
+        {
+            if (string.IsNullOrEmpty(caseName))
+                return;
+
+            // Arrange
+            var _signConfig = _protcolTestHelper.LoadSigningConfig().Result;
```
johnduhart, line 28: Don't use underscores for local variable names. Also, you're in an async method why are you calling `.Result`?

tests/Medidata.MAuth.Tests/xunit.runner.json
```diff
@@ -1,3 +1,5 @@
 ﻿{
-  "methodDisplay": "method"
+  "methodDisplay": "method",
+  "parallelizeTestCollections": false,
+  "maxParallelThreads": 1
```
johnduhart, line 4: Why was this required?

tests/Medidata.MAuth.Tests/MAuthProtocolSuiteTests.cs
```diff
@@ -0,0 +1,113 @@
...
+                Assert.Equal(sig, actualPayload);
+
+
+                // Verify string_to_sign is matched
+                // Act 
+                var result = await new MAuthCoreV2().GetSignature(request, authInfo);
+                var sts = await _protcolTestHelper.ReadStringToSign(caseName);
+
+                // Assert
+                Assert.Equal(sts, Encoding.UTF8.GetString(result));
```
johnduhart, line 75: What is the value of this specific test? Aren't we validating the signature and the string to sign is correct by verifying the headers. I'd rather not test internal details like this if we don't have to.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
```diff
@@ -188,33 +188,27 @@ public static string BuildEncodedQueryParams(this string queryString)
                 return string.Empty;
 
             var queryArray = queryString.Split('&');
+            var unescapedKeysAndValues = new KeyValuePair<string, string>[queryArray.Count()];
```
johnduhart, line 191: Just use `.Length` instead of `.Count()`

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
```diff
@@ -0,0 +1,131 @@
...
+            var completebodyFilePath = $"{_testCasePath}/{caseName}/{bodyFilepath}";
+            var bytes = File.ReadAllBytes(completebodyFilePath);
+            return Convert.ToBase64String(bytes);
+        }
+
+        public string[] GetTestCases()
+        {
+            var cases = Directory.Exists(_testCasePath)
+                ? Directory.GetDirectories(_testCasePath).Select(x => Path.GetFileName(x)).ToArray()
+                : new string[1];
```
johnduhart, line 80: `new string[1]` ?

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
prajon84: `Path` does not contain definition for `Join` in `net461`
and `Medidata.MAuth.Tests` has target frameworks of `net461` and `netcoreapp2.1`

tests/Medidata.MAuth.Tests/xunit.runner.json
prajon84: At some point, I thought it was required to parallelize tests. But, just verified its not needed. Will remove this. Thanks.

src/Medidata.MAuth.Core/MAuthCoreExtensions.cs
prajon84: Updated -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/9433f2479247bd7fd1e79d5a067a2b2b29a6da8e#diff-4819a82bb7be58721857c2b409e352acR191

tests/Medidata.MAuth.Tests/MAuthProtocolSuiteTests.cs
prajon84: Updated -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/9433f2479247bd7fd1e79d5a067a2b2b29a6da8e#diff-a00da24e99d915ff2dc2ddd89bda839cR28

tests/Medidata.MAuth.Tests/MAuthProtocolSuiteTests.cs
prajon84: Yes we are validating the headers too. Just following the `mauth-protocol-test-suite` test cases https://github.com/mdsol/mauth-protocol-test-suite/blob/develop/README.md#test-cases

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
prajon84: Updated to return `null` when no test cases found -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/2e7775cec007b04db1b8315b2556e8511427ad8a

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
johnduhart: Oh, so `Path.Combine` should be available then right?

prajon84: Yes. Doing that on the process of converting that.

tests/Medidata.MAuth.Tests/Infrastructure/MAuthServerHandler.cs
```diff
@@ -46,10 +59,11 @@ internal class MAuthServerHandler : HttpMessageHandler
                     {
                         security_token = new ApplicationInfo()
                         {
-                            Uuid = clientUuid,
+                            Uuid = isProtocolSuiteTest ? SigningAppUuid : clientUuid,
                             Name = "Medidata.MAuth.Tests",
                             CreationDate = new DateTimeOffset(2016, 8, 1, 0, 0, 0, TimeSpan.Zero),
-                            PublicKey = TestExtensions.ClientPublicKey
+                            PublicKey = isProtocolSuiteTest 
+                            ? SigningPublicKey : TestExtensions.ClientPublicKey
```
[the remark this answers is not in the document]
prajon84: Line `65` and `66` fitted in same line as length was under `120` -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/a022f854513263b6fed778cb4aba4614811d06aa#diff-7637ba2c5aebd0a96fd5e21714032388R65

tests/Medidata.MAuth.Tests/Infrastructure/TestExtensions.cs
```diff
@@ -123,5 +124,23 @@ public static HttpRequestMessage ToDefaultHttpRequestMessage(this RequestData da
...
+            PrivateKey = clientPrivateKey,
+            SignedTime = signedTime
+        };
+
+        public static HttpRequestMessage ToHttpRequestMessage(this UnSignedRequest data)
+        {
+            var result = new HttpRequestMessage(new HttpMethod(data.Verb), new Uri($"https://example.com{data.Url}"))
+            {
+                Content = !string.IsNullOrEmpty(data.Body) 
+                ? new ByteArrayContent(Convert.FromBase64String(data.Body)) : null,
```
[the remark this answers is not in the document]
prajon84: `StringContent` won't work as it needs code implementation change too.
Just indented the line -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/a022f854513263b6fed778cb4aba4614811d06aa#diff-2b827886e5236490770d5ec7b5ddbfdc

tests/Medidata.MAuth.Tests/Infrastructure/MAuthServerHandler.cs
```diff
@@ -16,6 +16,11 @@ internal class MAuthServerHandler : HttpMessageHandler
 
         public int SucceedAfterThisManyAttempts { get; set; } = 1;
 
+        private static ProtocolTestSuiteHelper _protocolSuiteHelper = new ProtocolTestSuiteHelper();
+
+        public static readonly string SigningPublicKey = _protocolSuiteHelper.GetPublicKey().Result;
+        public static readonly Guid SigningAppUuid = _protocolSuiteHelper.ReadSignInAppUuid().Result;
```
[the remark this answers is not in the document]
prajon84: Corrected as this -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/373eac3ef2e6e57c443867f052e6f4fe9d8dd423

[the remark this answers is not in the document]
hkurniawan-mdsol: That's not a Factory pattern though. The reason why we want a Factory pattern is when `MauthServerHandler` is instantiated, the `InitializeAsync` method should be called as well.

If `InitializeAsync` has to be called separately like 
`serverOptions.MAuthServerHandler = await new MAuthServerHandler().InitializeAsync(); `
a new dev wouldn't have a clue that it has to call `InitializeAsync` after instantiating `MAuthServerHandler`.

With the Factory pattern sample above (sorry missing private constructor)
```cs
internal sealed class MAuthServerHandler : HttpMessageHandler
{
    private MAuthServerHandler () { }
    ...
}
```
since you cannot instantiate `MAuthServerHandler` with `new` so the only way to do it is
```cs
serverOptions.MAuthServerHandler = await MAuthServerHandler.CreateAsync();
```
which would call `InitializeAsync` automatically.

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        public async Task<AuthenticationHeader> ReadAuthenticationHeader(string testCaseName)
+        {
+            var authzFilePath = Path.Combine(_testCasePath, testCaseName, $"{testCaseName}.authz");
+            return JsonConvert.DeserializeObject<AuthenticationHeader>(
+                await ReadValueFromPath(authzFilePath));
+        }
+
+        private async Task<UnSignedRequest> ReadRequestParameters(string requestPath) =>
```
hkurniawan-mdsol, line 103: Shouldn't this be `ReadUnsignedRequests`?

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
```diff
@@ -0,0 +1,129 @@
...
+        public async Task<string> ReadDigitalSignature(string testCaseName)
+        {
+            var sigFilePath = Path.Combine(_testCasePath, testCaseName, $"{testCaseName}.sig");
+            return await ReadValueFromPath(sigFilePath);
+        }
+
+        public async Task<AuthenticationHeader> ReadAuthenticationHeader(string testCaseName)
+        {
+            var authzFilePath = Path.Combine(_testCasePath, testCaseName, $"{testCaseName}.authz");
+            return JsonConvert.DeserializeObject<AuthenticationHeader>(
```
hkurniawan-mdsol, line 99: Why not use `System.Text.Json`?

tests/Medidata.MAuth.Tests/MAuthProtocolSuiteTests.cs
```diff
@@ -0,0 +1,122 @@
...
+                var data = new List<object[]>();
+                foreach (string item in cases)
+                {
+                    data.Add(new object[] { item });
+                }
+
+                return data;
+            }
+        }
+    }
```
hkurniawan-mdsol, line 121: [nits]
This can be refactored to
```cs
public static IEnumerable<object[]> TestCases =>
    new ProtocolTestSuiteHelper().GetTestCases()
    .Select(tc => new object[] { tc });
```

tests/Medidata.MAuth.Tests/MAuthProtocolSuiteTests.cs
```diff
@@ -0,0 +1,122 @@
...
+                // Act
+                var actualPayload = await new MAuthCoreV2().CalculatePayload(request, authInfo);
+
+                // Assert
+                Assert.Equal(sig, actualPayload);
+
+
+                // Verify string_to_sign is matched
+                // Act 
+                var result = await new MAuthCoreV2().GetSignature(request, authInfo);
```
hkurniawan-mdsol, line 72: Why `MAuthCoreV2` is instantiated again?

hkurniawan-mdsol:
Can we change this
https://github.com/mdsol/mauth-client-dotnet/blob/373eac3ef2e6e57c443867f052e6f4fe9d8dd423/tests/Medidata.MAuth.Tests/Infrastructure/AssertSigningHandler.cs#L32

To `return Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK));`

tests/Medidata.MAuth.Tests/Infrastructure/TestExtensions.cs
```diff
@@ -123,5 +124,23 @@ public static HttpRequestMessage ToDefaultHttpRequestMessage(this RequestData da
 
         public static HttpMethod ToHttpMethod(this string method) => new HttpMethod(method);
 
+        public static MAuthSigningOptions ProtocolTestClientOptions(Guid clientUuid, 
+            string clientPrivateKey, DateTimeOffset signedTime) => new MAuthSigningOptions()
```
hkurniawan-mdsol, line 128: Do we really need this extension method?
Looks like it's just called from 1 place.

tests/Medidata.MAuth.Tests/Infrastructure/TestExtensions.cs
```diff
@@ -123,5 +124,23 @@ public static HttpRequestMessage ToDefaultHttpRequestMessage(this RequestData da
...
+        {
+            ApplicationUuid = clientUuid,
+            PrivateKey = clientPrivateKey,
+            SignedTime = signedTime
+        };
+
+        public static HttpRequestMessage ToHttpRequestMessage(this UnSignedRequest data)
```
hkurniawan-mdsol, line 135: Same as this. Looks like this is just called from 1 place inside `MAuthProtocolSuiteTests`. Maybe just create a private method there?

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
```diff
@@ -0,0 +1,129 @@
...
+                : Path.GetFullPath(Path.Combine(currentDirectory, "../../../../../../mauth-protocol-test-suite"));
+            _testCasePath = Path.Combine(_testSuitePath, "protocols/MWSV2");
+        }
+
+        public async Task<SigningConfig> LoadSigningConfig()
+        {
+            var configFile = Path.Combine(_testSuitePath, "signing-config.json");
+
+            var signingConfig = await ReadSigningConfigParameters(configFile);
+            if (signingConfig is null )return null;
```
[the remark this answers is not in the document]
prajon84: No, its not possible. Therefore, when reading test cases if not found, then `case` is null then the test simply ends. 
But, this line was added as extra check before grabbing `privatekey`

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            var reqFilePath = Path.Combine(_testCasePath, testCaseName, $"{testCaseName}.req");
+            var unsignedRequest = await ReadRequestParameters(reqFilePath);
+            unsignedRequest.Body = !string.IsNullOrEmpty(unsignedRequest.Body) 
+                ? Convert.ToBase64String(unsignedRequest.Body.ToBytes()) : unsignedRequest.Body;
+
+            if (!string.IsNullOrEmpty(unsignedRequest.BodyFilePath))
+            {
+                unsignedRequest.Body = GetTestRequestBody(testCaseName, unsignedRequest.BodyFilePath);
+            }
```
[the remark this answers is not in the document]
prajon84: Oh, this is special case like as `binary-body` https://github.com/mdsol/mauth-protocol-test-suite/blob/develop/protocols/MWSV2/binary-body/binary-body.req
when there is no `body` but `body_filepath`. Therefore, we are trying to read the value of `body_filepath` which will be value of the body in this case.

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
```diff
@@ -0,0 +1,129 @@
...
+            return unsignedRequest;
+        }
+
+        public string GetTestRequestBody(string caseName, string bodyFilepath)
+        {
+            if (string.IsNullOrEmpty(bodyFilepath))
+                return null;
+            var completebodyFilePath = Path.Combine(_testCasePath, caseName, bodyFilepath);
+            var bytes = File.ReadAllBytes(completebodyFilePath);
```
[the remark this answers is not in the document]
prajon84: Yes, you are correct. `ReadAllBytesAsync` is not available in `.net461`.
But, just quickly tested the byte value received from`File.ReadAllBytes(filepath)` is not the same as:
`Encoding.UT8.GetBytes(await ReadValueFromPath(filePath))`.
If we want to read the value as `async`, then may be we can add a new method which read as bytes as:

```
 private async Task<byte[]> ReadBytes(string filePath)
        {
            using (FileStream stream = File.OpenRead(filePath))
            {
                byte[] result = new byte[stream.Length];
                await stream.ReadAsync(result, 0, (int)stream.Length);
                return result;
            }
        }
```
what do you think?

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
prajon84: Because, this solution uses `Newtonsoft.Json` at other multiple places too. Therefore, trying to keep the same.

tests/Medidata.MAuth.Tests/MAuthProtocolSuiteTests.cs
prajon84: Oh good catch. I will update it to instantiate at the top level only.

tests/Medidata.MAuth.Tests/Infrastructure/TestExtensions.cs
prajon84: Yes you are right. I will make it `private` and move it to `MAuthProtocolSuiteTests`

tests/Medidata.MAuth.Tests/Infrastructure/TestExtensions.cs
prajon84: Added as `private` -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/aa941fa38a149281e66d2b9321d9e2b93b93dcdf#diff-a00da24e99d915ff2dc2ddd89bda839cR119-R136

tests/Medidata.MAuth.Tests/MAuthProtocolSuiteTests.cs
prajon84: Refactored but kept the cases when `null` -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/aa941fa38a149281e66d2b9321d9e2b93b93dcdf#diff-a00da24e99d915ff2dc2ddd89bda839cL102-R115

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
prajon84: Updated -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/aa941fa38a149281e66d2b9321d9e2b93b93dcdf#diff-21ed0924b755dbaf0026fc8d3ae53e67R103

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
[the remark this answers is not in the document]
prajon84: Updated as I mentioned above -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/aa941fa38a149281e66d2b9321d9e2b93b93dcdf#diff-21ed0924b755dbaf0026fc8d3ae53e67R72

prajon84:
> Can we change this
> https://github.com/mdsol/mauth-client-dotnet/blob/373eac3ef2e6e57c443867f052e6f4fe9d8dd423/tests/Medidata.MAuth.Tests/Infrastructure/AssertSigningHandler.cs#L32
> 
> To `return Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK));`

@hkurniawan-mdsol -> Updated -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/aa941fa38a149281e66d2b9321d9e2b93b93dcdf#diff-bdbe3166d3470f433080680869fbe661

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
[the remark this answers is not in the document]
hkurniawan-mdsol: In that case we don't need to catch any exception that might occur when we read the file then.
https://github.com/mdsol/mauth-client-dotnet/blob/0c025c3edae6e359f6991bb2a0ee80ac86480c37/tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs#L122-L125

If there's something wrong reading the test cases/test files we should fail immediately.

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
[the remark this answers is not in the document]
hkurniawan-mdsol: Okay I see.

We can refactor it to be a bit more succinct like
```cs
unsignedRequest.Body = !string.IsNullOrEmpty(unsignedRequest.BodyFilePath) ?
    await GetTestRequestBody(testCaseName, unsignedRequest.BodyFilePath) :
    Convert.ToBase64String(unsignedRequest.Body.ToBytes());
```
If `string.IsNullOrEmpty(unsignedRequest)`, then the `ToBytes` will return `byte[0]` array and `ToBase64String` will convert that to empty string so we're good.

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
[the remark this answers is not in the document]
hkurniawan-mdsol: Oh cool! So I think we can replace all `ReadValueFromPath` with
```cs
Encoding.UTF8.GetString(await ReadBytes(...))
```

[nits] Maybe rename the method to `ReadAsBytes`.

tests/Medidata.MAuth.Tests/MAuthProtocolSuiteTests.cs
hkurniawan-mdsol: Do we need to check for the existence of the `test cases`?
If there are no `test cases`, what are we even testing?

tests/Medidata.MAuth.Tests/Infrastructure/MAuthServerHandler.cs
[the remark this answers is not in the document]
prajon84: Updated to factory pattern and updated all the instantiated places of `MAuthServerHandler` -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/a9b09ee1fe2aec6883859991a8150bb77ed22fb1

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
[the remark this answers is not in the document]
prajon84: Updated to use as `ReadAsBytes(string filePath)` -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/b1697d2e7aa062f9b627e6fdc70dfdfa811a6782#diff-21ed0924b755dbaf0026fc8d3ae53e67

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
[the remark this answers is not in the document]
prajon84: Updated -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/b1697d2e7aa062f9b627e6fdc70dfdfa811a6782#diff-21ed0924b755dbaf0026fc8d3ae53e67R54-L64

tests/Medidata.MAuth.Tests/MAuthProtocolSuiteTests.cs
prajon84: Well, since this test module `mauth-protocol-test-suite` is not being added to this `mauth-client-dotnet` yet unless it is made `public`. Therefore, `mauth-protocol-test-suite` is just cloned locally and when the unit test code finds it runs all 19 test cases. But, it does not find, currently, the unit test compiles for `empty caseName` which just ends the unit test as -> https://github.com/mdsol/mauth-client-dotnet/blob/feature/MCC-575820/tests/Medidata.MAuth.Tests/MAuthProtocolSuiteTests.cs#L29

But, if this part is removed to return empty list of object. And, if any contributor has not cloned the `mauth-protocol-test-suite` then it will fail.
Else have to return `Array.Empty<string>` in else case of `GetTestCases` but it will also yield me :
`System.InvalidOperationException : No data found for Medidata.MAuth.Tests.ProtocolTestSuite.MAuthProtocolSuiteTests.MAuth_Execute_ProtocolTestSuite`

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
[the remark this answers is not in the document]
prajon84: `Readsigningconfig` and `ReadPublicKey` are also consumed by `MAuthServerHandler`. Therefore, when "mauth-protocol-test-suite" is not loaded, then other unit tests should not be failing because of this. Therefore, added `try.. catch..` and return `null` for `DirectoryNotFoundException`. -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/57314249c8b57c4e0c2752a2972f31f8273c4199#diff-21ed0924b755dbaf0026fc8d3ae53e67

tests/Medidata.MAuth.Tests/MAuthProtocolSuiteTests.cs
hkurniawan-mdsol: So the question is, do we have to run those protocol test suites every time we make changes to mauth-client-dotnet?
If the answer is `yes` then we should fail if we can't read the test cases (due to not cloning the sub-module or other reasons).
Otherwise, maybe it's okay to return empty test data. 

The follow up question is, how do you know your changes are correct if you don't run the protocol test suites?

prajon84: > So the question is, do we have to run those protocol test suites every time we make changes to mauth-client-dotnet?
> If the answer is `yes` then we should fail if we can't read the test cases (due to not cloning the sub-module or other reasons).
> Otherwise, maybe it's okay to return empty test data.

Yes, we have to run protocol test suites if the changes are related to V2 since the protocol test suite is for V2. Since, `mauth-protocol-test-suite` is in the process of being public, we need to skip non "protocol-test-suite" test when just running other unit test. 
Once `mauth-protocol-test-suite` is public, then we do not need to skip the tests any more.
 
> The follow up question is, how do you know your changes are correct if you don't run the protocol test suites?

As mentioned above, we need to run the protocol test suites for the validation of the tests.

Therefore, I will update `CONTRIBUTING.md` how to run non `mauth-protocol-test-suite` test and `mauth-protocol-test-suite` test and remove the empty cases and handling empty cases too.

jfeltesse-mdsol: @hkurniawan-mdsol the stance for all the mauth implementations has been to wait until the suite is open sourced and run those tests locally if needs be, see https://mdsol.slack.com/archives/C2MDE04R5/p1594808785184000

hkurniawan-mdsol: @prajon84 @jfeltesse-mdsol 

I see the protocol test suites are very important to run against whenever we make changes. If we are shortcutting them due to not able to read the test cases, then changes can lead to a serious bug.

I see that currently they're not open sourced (but will eventually) so we can't add them as a sub-module and if you try to run the whole tests, some will fail. 
Which is fine (and good) since the dev is now aware of the test suites that they need to run against.
If we fail silently, this can be dangerous.

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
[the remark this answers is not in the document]
hkurniawan-mdsol: Okay. I think we need to remove the `try...catch` when we can have the protocol test suites as a sub-module.
In the mean time this is fine.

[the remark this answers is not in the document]
ykitamura-mdsol: maybe we can add a TODO comment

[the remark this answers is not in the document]
prajon84: Done -> https://github.com/mdsol/mauth-client-dotnet/pull/64/commits/8a4e91698bb12b5c02d9b1a00fd0c898f6705c27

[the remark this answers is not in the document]
hkurniawan-mdsol: Make sure the `TODO` comment is being tracked by the `Track List`.
https://docs.microsoft.com/en-us/visualstudio/ide/using-the-task-list?view=vs-2019#tokens-and-comments

jfeltesse-mdsol:
Merging EOD unless further comments

tests/Medidata.MAuth.Tests/ProtocolTestSuite/ProtocolTestSuiteHelper.cs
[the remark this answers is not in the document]
prajon84: image

I guess this is what you meant :)

PR #67, opened 2020-08-25
mention that the private key file is not cached

jfeltesse-mdsol:
Follow up of https://jira.mdsol.com/browse/MCC-655130

Open to suggestions to improve the copy (edit: _of the code comments I added_, this is not some "refactor the README" PR).

@dcassidy-mdsol @mdsol/libraries_dotnet @mdsol/architecture-enablement @mdsol/team-51

PR #68, opened 2020-11-17
[MCC-655130] Inflate private key upon set in options classes.

dhoizner:
Follow-up on #67.

Related to https://jira.mdsol.com/browse/MCC-655130.

Inflate private key upon set in the `_Options` classes in order to avoid reading it from disk every time a request is signed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changes in Medidata.MAuth
 
+## v5.0.1
+- **[Core]** Inflate private key upon set in options classes.
+
 ## v5.0.0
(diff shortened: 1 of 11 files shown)


prajon84 (commented):
Since there is code change, we need to update version, therefore we need update in `version.props` and `Changelog.MD`.

lschreck-mdsol (approved):
I think after updating the `version.props` this is good to go.

dhoizner:
@prajon84 @lschreck-mdsol updated `CHANGELOG.md` and `version.props`.

PR #69, opened 2020-12-14
[Housekeeping] Add Sync Method for MAuthSigningHandler

lyndon-gabriel:
The DPSS project is a synchronous process and leverages .NET 5 capability to send Synchronous HttpClient request. The MAuthSigningHandler currently does do not handle Send method. This PR adds the functionality to sign sync requests.

### DIFF
--- a/src/Medidata.MAuth.Core/IMAuthCore.cs
+++ b/src/Medidata.MAuth.Core/IMAuthCore.cs
@@ -5,17 +5,16 @@ namespace Medidata.MAuth.Core
 {
     internal interface IMAuthCore
     {
-        Task<HttpRequestMessage> SignAsync(HttpRequestMessage request, MAuthSigningOptions options);
+        Task<HttpRequestMessage> Sign(HttpRequestMessage request, MAuthSigningOptions options);
 
         bool Verify(byte[] signedData, byte[] signature, string publicKey);
 
-        Task<byte[]> GetSignatureAsync(HttpRequestMessage request, AuthenticationInfo authInfo);
+        Task<byte[]> GetSignature(HttpRequestMessage request, AuthenticationInfo authInfo);
 
         (string mAuthHeaderKey, string mAuthTimeHeaderKey) GetHeaderKeys();
 
 #if NET5_0
-
-        HttpRequestMessage Sign(HttpRequestMessage request, MAuthSigningOptions options);
+        HttpRequestMessage SignSync(HttpRequestMessage request, MAuthSigningOptions options);
 #endif
     }
 }
--- a/src/Medidata.MAuth.Core/IMAuthCore.cs
+++ b/src/Medidata.MAuth.Core/IMAuthCore.cs
@@ -5,12 +5,17 @@ namespace Medidata.MAuth.Core
 {
     internal interface IMAuthCore
     {
-        Task<HttpRequestMessage> Sign(HttpRequestMessage request, MAuthSigningOptions options);
+        Task<HttpRequestMessage> SignAsync(HttpRequestMessage request, MAuthSigningOptions options);
 
         bool Verify(byte[] signedData, byte[] signature, string publicKey);
 
-        Task<byte[]> GetSignature(HttpRequestMessage request, AuthenticationInfo authInfo);
+        Task<byte[]> GetSignatureAsync(HttpRequestMessage request, AuthenticationInfo authInfo);
 
         (string mAuthHeaderKey, string mAuthTimeHeaderKey) GetHeaderKeys();
+
+#if NET5_0
+
+        HttpRequestMessage Sign(HttpRequestMessage request, MAuthSigningOptions options);
+#endif
     }
 }
--- a/src/Medidata.MAuth.Core/MAuthCore.cs
+++ b/src/Medidata.MAuth.Core/MAuthCore.cs
@@ -6,7 +6,6 @@ using System.Threading.Tasks;
 using Org.BouncyCastle.Crypto.Engines;
 using Org.BouncyCastle.Security;
 using Medidata.MAuth.Core.Models;
-using System.IO;
 
 namespace Medidata.MAuth.Core
 {
@@ -26,7 +25,7 @@ namespace Medidata.MAuth.Core
             {
                 ApplicationUuid = options.ApplicationUuid,
                 SignedTime = options.SignedTime ?? DateTimeOffset.UtcNow,
-                PrivateKey = options.PrivateKey.Dereference().NormalizeLines()
+                PrivateKey = options.PrivateKey
             };
 
             var contents = await request.GetRequestContentAsBytesAsync().ConfigureAwait(false);
@@ -161,7 +160,7 @@ namespace Medidata.MAuth.Core
             {
                 ApplicationUuid = options.ApplicationUuid,
                 SignedTime = options.SignedTime ?? DateTimeOffset.UtcNow,
-                PrivateKey = options.PrivateKey.Dereference().NormalizeLines()
+                PrivateKey = options.PrivateKey
             };
 
             var contents = request.GetRequestContentAsBytes();
--- a/src/Medidata.MAuth.Core/MAuthCore.cs
+++ b/src/Medidata.MAuth.Core/MAuthCore.cs
@@ -20,16 +20,18 @@ namespace Medidata.MAuth.Core
         /// <returns>
         /// A Task object which will result the request signed with the authentication information when it completes.
         /// </returns>
-        public Task<HttpRequestMessage> Sign(HttpRequestMessage request, MAuthSigningOptions options)
+        public async Task<HttpRequestMessage> Sign(HttpRequestMessage request, MAuthSigningOptions options)
         {
-            return AddAuthenticationInfo(request, new PrivateKeyAuthenticationInfo()
+            var authInfo = new PrivateKeyAuthenticationInfo()
             {
                 ApplicationUuid = options.ApplicationUuid,
                 SignedTime = options.SignedTime ?? DateTimeOffset.UtcNow,
                 PrivateKey = options.PrivateKey.Dereference().NormalizeLines()
-            });
-        }
+            };
 
+            var contents = await request.GetRequestContentAsBytesAsync().ConfigureAwait(false);
+            return AddAuthenticationInfo(request, authInfo, contents);
+        }
 
         /// <summary>
         /// Verifies that the signed data is equal to the signature by using the public key of the keypair which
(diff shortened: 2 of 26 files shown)


prajon84:
I have few questions:

1. What is DPSS project ?
2. Why .Net 5 ?
3. Can it not use async http client ?

johnduhart (changes requested):
No. We can't just change method signatures like this and then bump the major version number. 

If it's come to the point that you need to make non-async MAuth signed HTTP calls within DPSS, I believe we've made a critical error with the design of DPSS.

@dhoizner

lyndon-gabriel:
> I have few questions:
> 
> 1. What is DPSS project ?
*DPSS is a project to isolated custom functions execution in a single process for the RaveEDC application. Custom functions are custom integration client write use Medidata.Core.Objects class library. This library is over 15 years old and those use the task based async patterns. We cannot modify the code based to use async because it will break millions of custom functions. One of the goal of the project is refactor Medida.Core.Objects to make it .NET Core compatible without any breaking changes.
Custom functions makes outbounds request.*

> 2. Why .Net 5 ?
*.NET has API's for synchronous  HttpClient request.*
One of the reason for updating tp .NET w
> 3. Can it not use async http client ?
*No for all the reasons mentioned above.*

lyndon-gabriel:
> No. We can't just change method signatures like this and then bump the major version number.
> 
> If it's come to the point that you need to make non-async MAuth signed HTTP calls within DPSS, I believe we've made a critical error with the design of DPSS.
> 
> @dhoizner

As mentioned in the previous comment, this is to maintain backward compatibility with Medidata.Core.Objects, the alternative is to use .Result, which you in the pass have pointed out leads to thread starvation. 

So if not going to change the API then what is your recommendation? @johnduhart

lyndon-gabriel:
Also its a minor signature change which follow the proper naming conventions for Async function and bumping major version is for exactly for changes like this, I don't understand why we can do this, please elaborate? @johnduhart

src/Medidata.MAuth.Core/IMAuthCore.cs
```diff
@@ -5,12 +5,17 @@ namespace Medidata.MAuth.Core
 {
     internal interface IMAuthCore
     {
-        Task<HttpRequestMessage> Sign(HttpRequestMessage request, MAuthSigningOptions options);
```
danielloganking, line 8: This is a breaking change. Why are we doing this way instead of making the new method `SignSync` (or some such).

src/Medidata.MAuth.Core/IMAuthCore.cs
```diff
@@ -5,12 +5,17 @@ namespace Medidata.MAuth.Core
...
 
-        Task<byte[]> GetSignature(HttpRequestMessage request, AuthenticationInfo authInfo);
+        Task<byte[]> GetSignatureAsync(HttpRequestMessage request, AuthenticationInfo authInfo);
 
         (string mAuthHeaderKey, string mAuthTimeHeaderKey) GetHeaderKeys();
+
+#if NET5_0
+
+        HttpRequestMessage Sign(HttpRequestMessage request, MAuthSigningOptions options);
+#endif
```
danielloganking, line 19: If we plan to add a synchronous implementation why limit it to `.NET 5`?

src/Medidata.MAuth.Core/MAuthCore.cs
```diff
@@ -124,5 +130,96 @@ internal async Task<string> CalculatePayload(HttpRequestMessage request, Private
         {
             return (Constants.MAuthHeaderKey, Constants.MAuthTimeHeaderKey);
         }
+
+#if NET5_0
```
danielloganking, line 134: Why are we adding these methods? Shouldn't the of the synchronous top level method just use AsyncToSync invocation so that under the hood we still have all the benefits of async and we don't have to maintain two versions of every method? Doesn't this violate DRY?

src/Medidata.MAuth.Core/MAuthCoreV2.cs
```diff
@@ -136,5 +137,115 @@ public async Task<byte[]> GetSignature(HttpRequestMessage request, Authenticatio
         {
             return (Constants.MAuthHeaderKeyV2, Constants.MAuthTimeHeaderKeyV2);
         }
+
+
+#if NET5_0
```
danielloganking, line 142: Same comment as on `MAuthCore`

src/Medidata.MAuth.Core/MAuthSigningHandler.cs
```diff
@@ -40,6 +40,7 @@ public MAuthSigningHandler(MAuthSigningOptions options, HttpMessageHandler inner
             _options = options;
         }
 
+
```
danielloganking, line 43: Nit: Whitespace

src/Medidata.MAuth.Core/MAuthSigningHandler.cs
```diff
@@ -59,13 +60,43 @@ public MAuthSigningHandler(MAuthSigningOptions options, HttpMessageHandler inner
...
             return await base
                 .SendAsync(request, cancellationToken)
                 .ConfigureAwait(continueOnCapturedContext: false);
         }
+
+
+#if NET5_0
+        /// <summary>
+        /// Signs an HTTP request with the MAuth-specific authentication information and sends the request to the
+        /// inner handler to send to the server as an asynchronous operation.
```
danielloganking, line 76: NIT: This is not asynchronous.

danielloganking (changes requested):
Concerns about the code:

1. We should not reimplement the entire stack in sync style and instead have the top level sync implementation use an [async to sync translation](https://docs.microsoft.com/en-us/archive/msdn-magazine/2015/july/async-programming-brownfield-async-development#the-blocking-hack) to keep as much of the code shared as possible and indicate the sync method is _non preferred_
2. If we implement this, we should implement this in a backwards compatible way
3. If we implement this, we should not only implement it for .NET5

However, all of the above is moot because I have fundamental concerns about even adding a synchronous version of an inherently async operation. Web calls are `async` in .NET full stop (and have been for since .NET 4.5). I am familiar with the changes in .NET5 and know of nothing that would necessitate this. If you have some other restriction requiring a web call in a synchronous context isn't that a client problem not a library problem? And if so, shouldn't the async-to-sync translation layer occur in the client? What makes this necessary as an addition to the library?

src/Medidata.MAuth.Core/MAuthCore.cs
lyndon-gabriel: Having a Sync function called async is consider a bad practice and leads to performance issues such as thread starvation.To avoid these issues the lower level methods need to synchronous.

https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md

src/Medidata.MAuth.Core/IMAuthCore.cs
lyndon-gabriel: So the normal convention in .NET is for Async functions to be Postfix with Async. To avoid breaking changes I can rename to SignSync but it seems odd and does follow the standard naming convention. But I can change if this is the main concern.

src/Medidata.MAuth.Core/IMAuthCore.cs
lyndon-gabriel: Good point. In general the synchronous in HttpClient was added in .NET 5, but the underlying functions does not need to be limited to .NET 5.

lyndon-gabriel:
> Concerns about the code:
> 
> 1. We should not reimplement the entire stack in sync style and instead have the top level sync implementation use an [async to sync translation](https://docs.microsoft.com/en-us/archive/msdn-magazine/2015/july/async-programming-brownfield-async-development#the-blocking-hack) to keep as much of the code shared as possible and indicate the sync method is _non preferred_
> 2. If we implement this, we should implement this in a backwards compatible way
> 3. If we implement this, we should not only implement it for .NET5
> 
> However, all of the above is moot because I have fundamental concerns about even adding a synchronous version of an inherently async operation. Web calls are `async` in .NET full stop (and have been for since .NET 4.5). I am familiar with the changes in .NET5 and know of nothing that would necessitate this. If you have some other restriction requiring a web call in a synchronous context isn't that a client problem not a library problem? And if so, shouldn't the async-to-sync translation layer occur in the client? What makes this necessary as an addition to the library?

So these changes will be used in RaveEDC application. In general this is a legacy code based and we can change the API surface to be asynchronous because it will be incompatible/break millions of Custom Functions. Custom Functions are written by Clients.

Sync over Async has major performance issues which have already caused CIRT's in production. It's not a scalable solution, all the underlining methods needs to be truly synchronous. However the async functions has its own benefits, which we should leverage when we can.

The changes in MAuthSigning handler will only work in .NET 5 or above. To changes in the underlying methods does not need to be limited to .NET .

src/Medidata.MAuth.Core/MAuthCore.cs
danielloganking: Considering [this is _exactly_ what the "sync" version of `Send` Microsoft added to HttpClient does](https://github.com/ManickaP/runtime/blob/7844fbf39812bd591dc10dc0fdfbd5fc1137d507/src/libraries/System.Net.Http/src/System/Net/Http/HttpClient.cs#L464-L470), I can assure you this the proper manner for async to sync translations. Yes, it is a kludge; because you are doing an inherently async operation within a sync context, it _must_ be.

src/Medidata.MAuth.Core/IMAuthCore.cs
danielloganking: The normal convention is when _adding_ `async` versions of existing synchronous methods is to append Async. Because most of the standard library API was defined prior to async it has this naming convention but for inherently async operations it is quite conventional to leave off the suffix.

And either way the point is moot because all existing clients depend on this method signature. Not making breaking changes is more important than conforming to a naming convention.

danielloganking:
> Sync over Async has major performance issues which have already caused CIRT's in production. It's not a scalable solution, all the underlining methods needs to be truly synchronous.

Considering the HttpClient method you are relying on already does this async to sync translation I don't see this argument holding water. No HttpClient call is truly synchronous. At some point it _always_ will translate from async to sync. How you do that can ameliorate some of the issues but yes, doing async things synchronously will always be more brittle than fixing the code to be `async`. I understand such a fix is not possible in a brownfield project like Rave (and I'm not suggesting you do that) but that doesn't mean we should abandon good practice in shared libraries.

src/Medidata.MAuth.Core/MAuthCore.cs
lyndon-gabriel: https://github.com/dotnet/runtime/blob/db58dcdb2a3276a61ce7f12b5d20c40426362556/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L38-L39

When sync is called it returns a ValueTask which different from a Task. ValueTask will complete synchronous if the underlying method is sync.

lyndon-gabriel:
> > Sync over Async has major performance issues which have already caused CIRT's in production. It's not a scalable solution, all the underlining methods needs to be truly synchronous.
> 
> Considering the HttpClient method you are relying on already does this async to sync translation I don't see this argument holding water. No HttpClient call is truly synchronous. At some point it _always_ will translate from async to sync. How you do that can ameliorate some of the issues but yes, doing async things synchronously will always be more brittle than fixing the code to be `async`. I understand such a fix is not possible in a brownfield project like Rave (and I'm not suggesting you do that) but that doesn't mean we should abandon good practice in shared libraries.


Im not suggesting we abandon the async functionality. This PR only adds logic to support the use case for sync http calls. I can make the adjusts to function names to be post fix with Sync and leave the original code as is. If we agree on this pattern.

danielloganking:
> I can make the adjusts to function names to be post fix with Sync and leave the original code as is. If we agree on this pattern.

This is definitely a better path forward but there is a still a concern with code reuse. We are duplicating a _lot_ of code here and I'm honestly not sure how we get around it. I would have to spend some time with the HttpClient implementation to fully grok the pattern but it seems like a minimum implementation uses value task to propagate results and a control bit (🤮) to indicate if the underlying call should be made synchronously or asynchronously?

To be clear I'm not against adding a truly sync option but we shouldn't do it by copying everything and doubling our possibility for bugs or worse the ability of the two implementation's logic to drift apart.

src/Medidata.MAuth.Core/IMAuthCore.cs
lyndon-gabriel: Okay so I reviewed and confirmed it needs to target .NET 5 or higher because the `HttpRequestMessage.ReadAsStream` is only available in .NET 5 or higher

lyndon-gabriel:
I address the comments, reverted the breaking changes. I Added the new methods with Sync postfix. Unfortunately the sync functionality needs to be limited to .NET 5 because it uses new API's add in DelegatingHandler and HttpMessage.

src/Medidata.MAuth.Core/MAuthCore.cs
```diff
@@ -124,5 +126,94 @@ internal async Task<string> CalculatePayload(HttpRequestMessage request, Private
         {
             return (Constants.MAuthHeaderKey, Constants.MAuthTimeHeaderKey);
         }
+
+ #if NET5_0
+        public HttpRequestMessage SignSync(HttpRequestMessage request, MAuthSigningOptions options)
```
danielloganking, line 131: As the `async` implementation has no `await`s can we do something with `ValueTask` to avoid duplicating the code? I know we want a fully sync operation but I can't help but feel that the code duplication is a bug waiting to happen.

src/Medidata.MAuth.Core/MAuthCore.cs
```diff
@@ -29,6 +30,7 @@ public Task<HttpRequestMessage> Sign(HttpRequestMessage request, MAuthSigningOpt
             });
         }
 
+
```
danielloganking, line 33: NIT: Whitespace

src/Medidata.MAuth.Core/MAuthCore.cs
lyndon-gabriel: So the task are generate from the .NET framework we just await. The only async part is reading the Request Body, I refactored the code to avoid the duplications

lyndon-gabriel:
@danielloganking all comments were addressed. Can we proceed with getting PR merged,?

src/Medidata.MAuth.Core/MAuthCore.cs
```diff
@@ -61,17 +65,36 @@ public bool Verify(byte[] signedData, byte[] signature, string publicKey)
...
+        /// The request which has the information (method, url and content) for the signature.
+        /// </param>
+        /// <param name="authInfo">
+        /// The <see cref="AuthenticationInfo"/> which holds the application uuid and the time of the signature.
+        /// </param>
+        /// <param name="requestContents">
+        /// The request body as a byte array.
+        /// </param>
+        /// <returns>A Task object which will result the SHA512 hash of the signature when it completes.</returns>
+        internal byte[] GetSignatureInternal(HttpRequestMessage request, AuthenticationInfo authInfo, byte[] requestContents)
```
danielloganking, line 85: NIT: `GetSignatureInternal` is a bit clunky, is there a name that more accurately reflects the purpose of this method? Perhaps `GenerateSignature`?

danielloganking (commented):
Can we add Unit Tests to `MAuthCoreTests` and `MAuthCoreV2Tests` that prove the signature is the same for sync and async calls? Also do we need to add sync versions of the existing tests?

src/Medidata.MAuth.Core/MAuthSigningHandler.cs
```diff
@@ -67,5 +67,34 @@ public MAuthSigningHandler(MAuthSigningOptions options, HttpMessageHandler inner
...
+#if NET5_0
+        /// <summary>
+        /// Signs an HTTP request with the MAuth-specific authentication information and sends the request to the
+        /// inner handler to send to the server a synchronous operation.
+        /// </summary>
+        /// <param name="request">The HTTP request message to sign and send to the server.</param>
+        /// <param name="cancellationToken">A cancellation token to cancel operation.</param>
+        /// <returns>Returns <see cref="Task{HttpResponseMessage}"/>. The task object representing the asynchronous
+        /// operation.</returns>
+        protected override HttpResponseMessage Send(
```
danielloganking, line 80: NIT: Should we add a `<remarks>` comment that the `Async` version should be preferred?

tests/Medidata.MAuth.Tests/MAuthAuthenticatorTests.cs
```diff
@@ -226,6 +254,55 @@ public static async Task AuthenticateRequest_WithValidMWSV2Request_WillAuthentic
...
+            var actual = mAuthCore.SignSync(testData.ToHttpRequestMessage(version), TestExtensions.ClientOptions(testData.SignedTime));
+
+            // Assert
+            Assert.Equal(expectedMAuthHeader, actual.Headers.GetFirstValueOrDefault<string>(Constants.MAuthHeaderKeyV2));
+            Assert.Equal(
+                testData.SignedTime.ToUnixTimeSeconds(),
+                actual.Headers.GetFirstValueOrDefault<long>(Constants.MAuthTimeHeaderKeyV2)
+            );
+        }
+#endif
```
danielloganking, line 304: Can we add a `SignRequest_WithMWSValidRequest_SignsIdenticallySyncAndAsync()` and ``SignRequest_WithMWSV2ValidRequest_SignsIdenticallySyncAndAsync()`?

lyndon-gabriel: There's already an identical test for the async version. I believe these are different scenario's and should be kept separate.

src/Medidata.MAuth.Core/MAuthSigningHandler.cs
lyndon-gabriel: So users will not directly interact with this class, you just register it via dependency injection. The method invoked on the HttpClient instance determines which method on the HttpHandler will be called  sync/async is called.

dhoizner: with a public class, i don't think we can guarantee that.

tests/Medidata.MAuth.Tests/MAuthAuthenticatorTests.cs
danielloganking: The point is to confirm that there is no functional drift between the two. Yes, we should test them separately and these cases are now covered. But it would also be a big if the two methods of signing produced different signatures. We should cover such a case tests.

lyndon-gabriel: Yeah so I agree but both test assert against the same static data. The signature is hard code in the JSON file and all test for both sync and async verifies it matches defined in the json. Im not sure how adding another test, to verify the same data will be helpful.

danielloganking (approved):
LGTM. I'd still like to see the nuts addressed and a equivalency test case but understand that since the tests are based on static data the is effectively tested. The desire for an explicit equivalency test is therefor also considered a nit and thus non-blocking.

@johnduhart @dhoizner please re-review and note any blockers.

lyndon-gabriel:
@johnduhart is there something you will like me to change or can we proceed to getting the PR merged. Thanks.

prajon84 (approved):
LGTM

johnduhart (approved):

PR #70, opened 2021-01-21
[MCC-721105] Avoid Multiple Outbound Requests for Application Info When Not Already Cached

dcassidy-mdsol:
### Summary of Issue

If the application info for a particular application uuid has not been cached in memory, it's possible for multiple threads to enter the cache factory method and send outbound requests for the data from the mAuth service.  When there is a high number of concurrent requests executing this logic, there can be significant blocking within the application until the requested info gets cached in memory.

The two concerns are:
- Multiple threads able to call the cache item factory method when application info is not already cached for a specific application uuid
- A new HttpClient (and associated resources) created for each outbound request

### Observed Behavior
During a restart of our application, requests get queued on the web server while the new process starts up.  Once the application has started, it gets the queued requests to process.  This can result in a higher than average number of requests starting to get processed by the application at the same time, before any application info has been cached for the application sending the requests.  This can result in many of the threads making outbound requests to the mAuth service for the same application info, which in turn results in a long period of blocking, followed by timeouts on many of the requests, all with the following call stack:

Screen Shot 2021-01-21 at 11 13 09 AM

Once the application info is cached in memory, the application recovers and behaves as expected.  We've observed the blocking last as long as 10-15 minutes before the application recovers.

#### Restrict to a single outbound request per application uuid not already cached in memory
- Store the request retrier Task as a Lazy object in the MemoryCache to defer the start of the outbound request until after the request retrier Task has been placed in the cache.
- Implement a version of the GetOrCreate cache extension method that locks based on the cache key, to limit the calls to the cache item factory method to a single thread per key not already in the cache.

#### Reduce number of HttpClient objects (and associated resources) created
- Creation of the HttpClient has been moved up one level, to the MAuthAuthenticator, and that client is passed to each new MAuthRequestRetrier.  The HttpClient is held in a Lazy object, to defer creation until it's needed.

I've reproduced the issue at a smaller scale in our sandbox environment, and no longer see the performance impact with these changes in place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 ## v5.1.0
 - **[Core]** Added multi-target for .NET 5 to support synchronus HttpClient requests.  
 - **[Core]** Updated MAuthSigningHandler to sign synchronus requests.  
+- **[Core]** Updated MAuthAuthenticator to create and reuse a single HttpClient instead of creating a new one for each MAuthRequestRetrier.
+- **[Core]** Updated MAuthAuthenticator to limit the calls to the cache item factory method to a single thread per key not already in the cache.
 
 ## v5.0.1
 - **[Core]** Inflate private key upon set in options classes.
--- a/src/Medidata.MAuth.Core/AsyncLazy.cs
+++ b/src/Medidata.MAuth.Core/AsyncLazy.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace Medidata.MAuth.Core
+{
+    public class AsyncLazy<T> : Lazy<Task<T>>
+    {
+        public AsyncLazy(Func<T> valueFactory)
+            : base(() => Task.Factory.StartNew(valueFactory))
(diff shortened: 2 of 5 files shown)


dcassidy-mdsol:
@prajon84 @hkurniawan-mdsol @lschreck-mdsol Could I get this PR reviewed?  We have a recurring issue in production related to this, and we'd like to include it in our R3 release.

prajon84:
Can you update `changelog` too ?

jfeltesse-mdsol:
Is v5.1.0 published already? Because if it is, a version bump would be needed.

lschreck-mdsol (approved):

prajon84:
> Is v5.1.0 published already? Because if it is, a version bump would be needed.

No, v5.1.0 has not been published yet. So, version bump is not needed.

prajon84 (approved):

dcassidy-mdsol:
@lschreck-mdsol @prajon84 Thanks for the speedy review, I appreciate it!

art2003 (approved):

lyndon-gabriel (approved):

PR #78, opened 2021-05-14
[MCC-771288] Add Workflow for Publishing the Packages

lschreck-mdsol:
This PR adds a new workflow to publish the packages to our Artifactory feed.

@mdsol/enablement-dotnet Please review and merge - thanks!

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,4 +1,4 @@
-name: Build and Test
+name: Build, Test and Publish
 on:
   push:
     branches:
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,8 +1,8 @@
 name: Build, Test and Publish
 on:
   push:
-    branches:
-      - develop
+    tags:
+      - 'release/*'
 jobs:
   Publish:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


jfeltesse-mdsol:
what's the trigger? every time there's a push to develop? preferrably, publishing runs only upon tagging (a version can be made of several PRs)

lschreck-mdsol:
> what's the trigger? every time there's a push to develop? preferrably, publishing runs only upon tagging (a version can be made of several PRs)

Yes, currently if `develop` gets a merge - I think we can follow the "one PR - one version" cadence for libraries - especially because I don't think we have any big feature release in the future (or if we have, we can change the configuration of the publish workflow).

jfeltesse-mdsol:
Thanks. Also, this is going to be very convenient to re-tag the old releases after 5.0.0 so they get published now. Will do once this PR is in!

lschreck-mdsol:
Ok, and I merged one PR which removes a dependency so I bumped the version number here.

jfeltesse-mdsol (approved):

jfeltesse-mdsol:
Please make a develop -> master PR and then we can tag off master.

About my previous comment, alas, that won't apply since old commits wouldn't yet feature the GHA test/publish flow... 🤦

PR #82, opened 2021-05-26
[MCC-772940] Reorganize for compatibility

lschreck-mdsol:
This PR (almost entirely) eliminates the library multitargeting and sets every library to the closest target framework as possible without losing functionality.

In order to be simpler and straightforward, the unit test project also got separated by target framework/library.

Currently NRTM until the Artifactory credentials getting sorted out and a Windows-based workflow added for running the classic .NET Framework specific tests.

@mdsol/enablement-dotnet


Jman420 (changes requested):
This PR contains secrets which must be invalidated.

prajon84:
cc: @lyndon-gabriel

version.props
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project>
   <PropertyGroup>
-    <Version>5.1.2</Version>
+    <Version>5.1.3</Version>
```
prajon84, line 4: Since `version` is bumped. Needs `CHANGELOG.md` also update ?

lschreck-mdsol: You are right, added d4013aa

src/Medidata.MAuth.Core/HttpRequestMessageExtensions.cs
```diff
@@ -14,18 +14,15 @@ public async static Task<byte[]> GetRequestContentAsBytesAsync(this HttpRequestM
...
+            using var memoryStream = new MemoryStream();
+
+            if (request.Content != null)
             {
-                if (request.Content != null)
-                {
-                    request.Content.ReadAsStream().CopyTo(memoryStream);
-                }
-                return memoryStream.ToArray();
+                request.Content.ReadAsStreamAsync().GetAwaiter().GetResult().CopyTo(memoryStream);
```
lyndon-gabriel, line 23: This can cause performance issues. The Sync API was introduced in .NET 5 which is why we multi targeted. AFAIK This will allocate 2 threads which was caused major performance issue for RaveEDC in the past.

lschreck-mdsol: As covering .NET Standard 2.0 (i.e. .NET Framework 4.6.1) and .NET 5.0 at the same time supposedly won't cause any dependency issues (for now) I restored the multitargeting for this case. 9c0e5c6

src/Medidata.MAuth.Core/MAuthCore.cs
```diff
@@ -166,8 +166,6 @@ public HttpRequestMessage SignSync(HttpRequestMessage request, MAuthSigningOptio
             var contents = request.GetRequestContentAsBytes();
             return AddAuthenticationInfo(request, authInfo, contents);
         }
-
 #endif
-
     }
 }
```
prajon84, line 171: new line

tests/Directory.Build.props
```diff
@@ -0,0 +1,25 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.9.4" />
+    <PackageReference Include="Moq" Version="4.16.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="xunit" Version="2.4.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
+</Project>
```
prajon84, line 25: New line.

jfeltesse-mdsol (commented):
The artifactory secrets should be set but instead of a "password", the api key should have been set as `ARTIFACTORY_API_KEY`.

Jman420 (approved):
Updated file names are sufficient.  Thanks guys, looks good to me.

.github/workflows/publish.yaml
```diff
@@ -13,10 +13,8 @@ jobs:
           submodules: 'recursive'
       - name: Build the packages
         run: dotnet build --configuration Release /p:PackageOutputPath=$GITHUB_WORKSPACE/artifacts
-      - name: Run the tests
-        run: dotnet test $GITHUB_WORKSPACE/Medidata.MAuth.sln --framework net5.0
       - name: Publish the packages
         run: dotnet nuget push "$GITHUB_WORKSPACE/artifacts/*.nupkg" --source $ARTIFACTORY_PUSH_TARGET --api-key $API_KEY
         env:
-          API_KEY: ${{ format('{0}:{1}', secrets.ARTIFACTORY_USERNAME, secrets.ARTIFACTORY_PASSWORD) }}
+          API_KEY: ${{ format('{0}:{1}', secrets.ARTIFACTORY_USERNAME, secrets.ARTIFACTORY_API_KEY) }}
```
rkorneev, line 19: We have org level secret `ARTIFACTORY_PASSWORD` (paired with `ARTIFACTORY_USERNAME`), but not `ARTIFACTORY_API_KEY`.
So `ARTIFACTORY_API_KEY` is not going to work unless you override on repo level

jfeltesse-mdsol: That's what we thought and the reason we had artifactory password, like the other repos using GHA in our org. Turns out, org level secrets are only set for private repos. 
We had a long thread on Slack about that but bottom line, is we created a new artifactory account and had the credentials set for username and api key, hence the changes here.

src/Medidata.MAuth.Core/HttpRequestMessageExtensions.cs
```diff
@@ -17,14 +17,13 @@ public async static Task<byte[]> GetRequestContentAsBytesAsync(this HttpRequestM
 #if NET5_0
         public static byte[] GetRequestContentAsBytes(this HttpRequestMessage request)
         {
-            using (var memoryStream = new MemoryStream())
+            using var memoryStream = new MemoryStream();
```
danielloganking, line 20: Doesn't this require us to specify the LangVersion as 8 or greater as "bracket-less using statements" is a feature of C# not of any particular .NET version?

[Relevant doc](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement)

src/Medidata.MAuth.Core/MAuthCoreV2.cs
```diff
@@ -148,7 +148,7 @@ internal byte[] GenerateSignature(HttpRequestMessage request, AuthenticationInfo
         internal string CalculatePayload(
             HttpRequestMessage request, PrivateKeyAuthenticationInfo authInfo, byte[] requestContents)
         {
-            
+
```
danielloganking, line 151: NIT: no need for new line to start the method

tests/Medidata.MAuth.Tests.Common/Infrastructure/AssertSigningHandler.cs
```diff
@@ -31,19 +31,5 @@ internal class AssertSigningHandler : HttpMessageHandler
...
-        {
-            MAuthHeaderV2 = request.Headers.GetFirstValueOrDefault<string>(Constants.MAuthHeaderKeyV2);
-            MAuthHeader = request.Headers.GetFirstValueOrDefault<string>(Constants.MAuthHeaderKey);
-
-            MAuthTimeHeaderV2 = request.Headers.GetFirstValueOrDefault<string>(Constants.MAuthTimeHeaderKeyV2);
-            MAuthTimeHeader = request.Headers.GetFirstValueOrDefault<string>(Constants.MAuthTimeHeaderKey);
-
-            return new HttpResponseMessage(HttpStatusCode.OK);
-        }
-#endif
```
danielloganking, line 47: Is this no longer needed for synchronous sending support?

.github/workflows/publish.yaml
rkorneev: I see, thank you for explaining that.

tests/Medidata.MAuth.CoreTests/MAuthAuthenticatorTests.cs
```diff
@@ -446,12 +445,6 @@ public static async Task AuthenticateRequest_WithDefaultRequest_WhenV2Fails_Fall
 
             // Assert
             Assert.True(isAuthenticated);
-            mockLogger.Verify(x => x.Log(
-                    LogLevel.Warning, It.IsAny<EventId>(),
-                    It.Is<FormattedLogValues>(v => v.ToString()
-                        .Contains("Completed successful authentication attempt after fallback to V1")),
-                    It.IsAny<Exception>(), It.IsAny<Func<object, Exception, string>>()
-                ));
```
danielloganking, line 454: Why are we removing this check?

src/Medidata.MAuth.Core/HttpRequestMessageExtensions.cs
lschreck-mdsol: The `LangVersion` is set to `latest` [here](https://github.com/mdsol/mauth-client-dotnet/blob/158f171de75c2a06c17e132c2591b05751b50652/Directory.Build.props#L7) which is C#9 currently so this is OK.

tests/Medidata.MAuth.Tests.Common/Infrastructure/AssertSigningHandler.cs
lschreck-mdsol: We don't have `Send()` method anymore in the `HttpMessageHandler` under .NET 5.0 and `SendAsync()` can handle the tests without issues so I removed this method.

tests/Medidata.MAuth.CoreTests/MAuthAuthenticatorTests.cs
lschreck-mdsol: Unfortunately `FormattedLogValues` is not accessible from the outside anymore in .NET 5.0 (I believe MS changed its visibility to `internal`) so I could not find any sensible way to verify the log message.

src/Medidata.MAuth.Core/HttpRequestMessageExtensions.cs
danielloganking: 👍 I looked for this but missed it 🤦‍♂️

tests/Medidata.MAuth.Tests.Common/Infrastructure/AssertSigningHandler.cs
danielloganking: I know that Rave depends on the sync sending but don't really know the details of their needs. Might help to request @lyndon-gabriel to give this a look see. That said if the Sync tests pass it should be fine? 🤷‍♂️

lschreck-mdsol: Yeah, this handler is just asserting the given request so does not contribute to the test act (that is, semantically it's not part of the system under test).

jfeltesse-mdsol:
Looks like the things that have been pointed out have either been addressed or they can't be fixed so merging for now.

PR #84, opened 2021-06-23
create extension for aspnet core objects

yongliu-mdsol:
Issue:

We often need the extract app uuid from the mauth header to check/authorize the app's permission.
Currently, the mauth Core Extensions only supports HttpRequestMessage. To get around it, we've been convert it to HttpRequest object. The method that converts HttpRequestMessage to HttpRequest is already in Medidata.MAuth.AspNetCore.

To reduce duplicated code from our projects, I'm adding the extension to support HttpRequest.


danielloganking (approved):
LGTM. But maybe we should investigate storing the authenticated identity in a [Request Feature](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/request-features?view=aspnetcore-5.0) so clients don't need to re-parse the header if they want access to the identity?

johnduhart:
> But maybe we should investigate storing the authenticated identity in a Request Feature so clients don't need to re-parse the header if they want access to the identity?

I'm going to strongly suggest that we close this PR in favor of implementing exactly that. Converting an HttpRequest to an HttpRequestMessage isn't a free operation, and doing so to just read a header seems wasteful. I don't feel good about including a that as an API in the main library.

yongliu-mdsol:
Agreed. This is definitely not the ideal approach.
The middleware in aspnetcore already converts the object to  HttpRequestMessage and extract the authHeaderValue.
What we really need is to get the app uuid from authenticated request.

However, I don't feel like it's a small change to implement with request feature. It seems to involve with some refactoring work in the Mauth.Core as well.

I can close this PR if @mdsol/architecture-enablement can help implementing a better way for clients to retrieve app uuid without re-parse the header.

jfeltesse-mdsol:
> I can close this PR if @mdsol/architecture-enablement can help implementing a better way for clients to retrieve app uuid without re-parse the header.

When you say "can help", what exactly do you have in mind?

FWIW Æ doesn't own mauth-client-dotnet, indeed we've been contributing lately, but at the end of the day this package is "owned" by the _entire_ .NET community at Medidata.

yongliu-mdsol:
@jfeltesse-mdsol I was under the impression that AE team owns it and the dotnet community can contribute. If that's not the case, it's fine. 

I was referring to the changes that @danielloganking and @johnduhart suggested.

PR #96, opened 2023-03-21
Add time constraint between signature and verification

eraffel-MDSol:


src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -109,10 +114,22 @@ private async Task<bool> Authenticate(HttpRequestMessage request, MAuthVersion v
                     authInfo.ApplicationUuid.ToString(),
                     () => SendApplicationInfoRequest(authInfo.ApplicationUuid)).ConfigureAwait(false);
 
+            if (!IsSignatureTimeValid(authInfo.SignedTime))
+            {
+                return false;
+            }
```
johnduhart, line 120: In my Go implementation I put this check before the public key lookup given it's a cheap check.

johnduhart:
@mdsol/enablement-dotnet

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
eraffel-MDSol: https://github.com/mdsol/mauth-client-dotnet/pull/96/commits/9ac0976695bf4802e41d9f271211381e259a5456

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -104,15 +109,27 @@ private async Task<bool> Authenticate(HttpRequestMessage request, MAuthVersion v
 
             var mAuthCore = MAuthCoreFactory.Instantiate(version);
             var authInfo = GetAuthenticationInfo(request, mAuthCore);
-            
+
+            if (!IsSignatureTimeValid(authInfo.SignedTime))
+            {
+                return false;
```
prajon84, line 115: Do we need to throw `TimeOutException` and log error similar to other cases ?

awiesendangermdsol: How about a log though? If I had to debug why? Not sure how logging is throughout this library.

prajon84: Yeah at least a logger would be beneficial too.

Jman420: Agreed on logging.

Jman420 (changes requested):

src/Medidata.MAuth.Core/MAuthAuthenticator.cs
eraffel-MDSol: https://github.com/mdsol/mauth-client-dotnet/pull/96/commits/bf7b84829178af0a6d14df0fe5ed18f5518c751d

Jman420 (approved):

PR #99, opened 2023-04-03
Converting logs "information" to "debug"

prajon84:
Can we move these `information` logs to `debug` ?

What do you think?
@mdsol/enablement-dotnet 
@jcarres-mdsol

### DIFF
--- a/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
+++ b/src/Medidata.MAuth.Core/MAuthAuthenticator.cs
@@ -51,7 +51,7 @@ namespace Medidata.MAuth.Core
         {
             try
             {
-                _logger.LogInformation("Initiating Authentication of the request.");
+                _logger.LogDebug("Initiating Authentication of the request.");
 
                 var authHeader = request.GetAuthHeaderValue();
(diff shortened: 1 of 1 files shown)


src/Medidata.MAuth.Core/MAuthAuthenticator.cs
```diff
@@ -103,9 +103,12 @@ public async Task<bool> AuthenticateRequest(HttpRequestMessage request)
 
         private async Task<bool> Authenticate(HttpRequestMessage request, MAuthVersion version, Guid signedAppUuid)
         {
-            var logMessage = "Mauth-client attempting to authenticate request from app with mauth app uuid " +
-                $"{signedAppUuid} to app with mauth app uuid {_options.ApplicationUuid} using version {version}";
-            _logger.LogInformation(logMessage);
+            _logger.LogDebug(
```
jcarres-mdsol, line 106: This is the only one that maybe we want to keep, I'd like to hear from @danielloganking  , @johnduhart  or other dotnet practicioners

prajon84: May be @lyndon-gabriel , @dcassidy-mdsol  has to say anything ?

